PR #23, opened 2019-10-11
chore: bump node mod deps

vobu:


petermuessig:
Will be handled with:
https://github.com/petermuessig/ui5-ecosystem-showcase/pull/24

PR #36, opened 2019-10-25
feat(server): reuse middleware w/ different mountpoint

vobu:
just FTR: had to 
`find . -type d -name node_modules -delete`
then
`yarn`
again to get everything working.
a simple `yarn` after the package bumps didn't do it. 
*shrug*


petermuessig:
Maybe we add the yarn autoclean command: https://yarnpkg.com/lang/en/docs/cli/autoclean/

.gitignore
```diff
@@ -58,3 +58,6 @@ typings/
 
 # UI5 tooling dist folder
 dist
+
+# WebStorm/IntelliJ IDE
+.idea
```
petermuessig: näääh :-)

petermuessig: But well, ok, let's keep it to avoid polluting the repo by webstorm users... *lol*

Issue #115, opened 2020-02-05
With transpiled files, the following error comes while deploying: File contains different line feeds. Mixed line feeds are not supported


arindamseth:
With transpiled files, the following error comes while deploying: 
"File contains different line feeds. Mixed line feeds are not supported"
Is it possible to follow existing EOL characters while transpiling?

vobu:
hi, hm, not sure, never had this case. would you mind switching to unix-stlye line feeds (`lf`) and converting the files in question to that linefeed-style?

arindamseth:
Yes, as a measure to use this, I have switched to LF instead of CRLF. However I was wondering if it was possible for this tool to follow existing EOL characters. Thanks

vobu:
don't see this (mixed line feeds) being done - I'd rather use a .editorrc or .prettierrc forcing `lf` linefeeds :)

lboehm:
Just my two cents :) I had the same problem, and first I forced my editor (on windows) to use unix style linefeeds. But I was not satisfied with this approach (I don't remember why 🤦‍♂), so I now modify linefeed when transpiling:

````
var correctLinefeed = result.code.replace(/\r\n|\r|\n/g, '\r\n');
resource.setString(correctLinefeed);
workspace.write(resource);
````
(*This snippet has to be put in promises `then` in [transpile module](https://github.com/petermuessig/ui5-ecosystem-showcase/blob/master/packages/ui5-task-transpile/lib/transpile.js)*)

This code changes the linefeed to `\r\n`, it would be better to check which OS is used and set the linefeed based on the OS. But as we all are using Windows this works fine for us.

HTH

vobu:
cool, would you put this in a PR?

lboehm:
Okay, I'll give it a try :)

PR #126, opened 2020-02-16
Fix mixed linefeeds

lboehm:
As discussed in Issue #115 I've opened this pull request to fix the issue with mixed line feeds. As this issue is only relevant for deployment I've added the code only in the `transpile` task and not also in `live-transpile` task. 

If you think, I should add it to `live-transpile` task, let me know.


vobu:
cool shakes, thx already! and as you suggest, would you mind including in `live-transpile` as well. thanks!

lboehm:
Yes, I'll add it on the weekend

lboehm:
Hey :) I've added the commit for live transpile.

vobu:
very cool, thx for your contribution!

PR #130, opened 2020-02-25
Revert back UI5 version to 1.73.1 due to bug in >=1.74 with missing resources

petermuessig:
…esources

### DIFF
--- a/packages/ui5-app/package.json
+++ b/packages/ui5-app/package.json
@@ -29,10 +29,10 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


petermuessig:
@openui5/sap.ui.core doesn't include namespace sap/ui/test in the NPM package and therefore OpenUI5 applications using >=1.74 will have build and runtime issues. Issue will be addressed asap and we will provide a fix soon. Rolling back to 1.73.1 version for the show-case repo now.

PR #142, opened 2020-03-13
Fix bug in ui5-middleware-cfdestination

IObert:


vobu:
LGTM - even though it sits (git-wise) on top your ui5-task-tipper :)
if you want, re-write history and update, please (if not, i'll just patch this through and you'll be forever in the (git) history of this :) )

IObert:
I noticed that I forgot to fetch the latest commits when it was too late 😅
I'm good with whatever option you prefer

IObert:
Thanks for merging. Do you already know when you'll publish the next version?

PR #154, opened 2020-03-26
Correct prop name

IObert:


petermuessig:
Ah, Marius - thanks for the hint - but I would prefer to fix the name of the configuration property back to path due to compatibility reasons. Maybe I just support both.

petermuessig:
I think this was a mistake by @vobu and I have overseen this in my review. I created a PR now which fixes this, by adding the option to have both properties available for usage: `watchPath` and `watch`!

https://github.com/petermuessig/ui5-ecosystem-showcase/pull/156

I will close your PR since you need to rebase nevertheless - you set up your change on a very old commit.

PR #156, opened 2020-03-27
Fix the livereload watchPath config

petermuessig:


petermuessig:
@IObert - thanks for the hint - this will fix the issue - I will release the new version asap

PR #163, opened 2020-04-02
Override security settings to avoid failure or the approuter on start

IObert:
It would be even better if I would use `xsenv` first to check whether there is a UAA but this is the hot fix. I hope that works out for you :)

### DIFF
--- a/packages/ui5-middleware-cfdestination/lib/cfdestination.js
+++ b/packages/ui5-middleware-cfdestination/lib/cfdestination.js
@@ -18,15 +18,19 @@ const log = require("@ui5/logger").getLogger("server:custommiddleware:cfdestinat
  * @param {string} [parameters.options.configuration] Custom server middleware configuration if given in ui5.yaml
  * @returns {function} Middleware function to use
  */
-module.exports = function ({resources, options}) {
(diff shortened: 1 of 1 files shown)


vobu:
hmmm, i get the ida behind the PR - but wouldn't that already be possible by including `authenticationType` per route à la:
```json
xsappConfig.authenticationMethod = "route"
"routes": [
    {
      "authenticationType": "none",
      "source": "^/backend/(.*)$",
      "destination": "backend",
      "target": "$1"
    }, 
    {
      "authenticationType": "basic",
      "source": "^/other/(.*)$",
      "destination": "other",
      "target": "$1"
    },
   {
      "authenticationType": "xsuaa",
      "identityProvider": "some_idp"
      "source": "^/yabe/(.*)$",
      "destination": "someDest",
      "target": "$1"
    }
]
```

IObert:
The idea is that the embedded approuter (the one that runs locally) will ignore all auth mechanisms. Even when you choose `xsappConfig.authenticationMethod = "route"` and only define the routes, the approuter will try to read the env var during startup - and fail if they are missing.

vobu:
ah, check check. Will add your explanation as a src comment, then merge.

Issue #174, opened 2020-04-13
General configuration from environment variables and .env files


esjewett:
It would help a lot for CI scenarios if settings could optionally be pulled from environment variables. Support for dot-env on top of this would be fantastic. I'm mostly interested in this for ui5-middleware-simpleproxy in order to set the `baseUri`, but seems like something that would be a good standard for all tasks and middleware. Thoughts? Would pull requests to implement this be accepted in theory?

vobu:
absolutely. dotenv rocks 👍🏻

petermuessig:
@esjewett - yes, pull requests will be appreciated! :-)

petermuessig:
PR #180 submitted - thanks for your contribution... :-)

PR #179, opened 2020-04-16
New option in ui5-middleware-servestatic to use paths from .env file

stockbal:
Hi @petermuessig,

I added a new config option called `pathIsEnvVar`. If this option is set to `true` it is expected that the `rootPath` variable contains an environment variable contained in a `.env` file.

This makes collaberation in a ui5 repository easier as the `ui5.yaml` does not have to be adjusted by every user locally - not all users store their static resources in the same path.

Regards
Ludwig

### DIFF
--- a/packages/ui5-middleware-servestatic/README.md
+++ b/packages/ui5-middleware-servestatic/README.md
@@ -54,8 +54,7 @@ server:
     afterMiddleware: compression
     mountPath: /resources
     configuration:
-      rootPath: SAPUI5_SDK_1_60__RESOURCES
-      pathIsEnvVar: true
+      rootPath: ${env.SAPUI5_SDK_1_60__RESOURCES}
 ```
 ## How it works
 
--- a/packages/ui5-middleware-servestatic/README.md
+++ b/packages/ui5-middleware-servestatic/README.md
@@ -34,7 +34,7 @@ npm install ui5-middleware-servestatic --save-dev
 
 > As the devDependencies are not recognized by the UI5 tooling, they need to be listed in the `ui5 > dependencies` array. In addition, once using the `ui5 > dependencies` array you need to list all UI5 tooling relevant dependencies.
 
-2. configure it in `$yourapp/ui5.yaml`:
+2. configure it in `$yourapp/ui5.yaml`:  
 
 ```yaml
 server:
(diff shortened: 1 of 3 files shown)


petermuessig:
Hi @stockbal,

thanks for this contribution. This is really interesting as @vobu and I had a similar discussion with @esjewett in the issue #174 . We could enhance the middlewares to be able to consider the properties in general from `.env` via a special syntax, e.g. something like this:

```
     rootPath: ${env.myRootPath}
```

The identifier for the env variable would be: ${env.XXX} and can be identified with the following regexp: `^\$\{env\.(.*)\}$`. What do you think about this? 

BR,
Peter

stockbal:
Hi @petermuessig ,

sounds like a cool solution. 

Would you propose to do the parsing of the config option value in each middleware/task anew or maybe put the logic in a core npm package so we can just call a function `getConfigOption(options.configuration.rootPath)` to get the concrete value behind a config option?

Regards,
Ludwig

petermuessig:
Hi @stockbal ,

thanks for the adoption of your PR. I also thought about centralizing this logic in a core module - maybe as part of the repository. But we can even do this in a second step. 

BR,
Peter

petermuessig:
But even rebase and merge is fine - your commits look ok.

esjewett:
Late to the party, but why not `${XXX}` instead? My preference would be to introduce a syntax-free option until we can consider this a little more and introduce something that is in line with existing syntax for specifying environment variables in YAML.

RandomByte:
> with existing syntax for specifying environment variables in YAML

Do you have any references? I wondered the same earlier but couldn't find anything specific for YAML and environment variables in a quick search.

I'm thinking that with the `env.` prefix we can allow some other variables later on. Like project specific information from the UI5 tooling on `project.` (e.g. `project.namespace`).

esjewett:
I've not done a ton of YAML configuration file work, but a few options I've seen for the value-side of a YAML parameter:

- `env(MY_VARIABLE)`
- `${MY_VARIABLE)`
- `$${MY_VARIABLE}` (why? 😂)
- `!ENV ${MY_VARIABLE}`
- `${env.MY_VARIABLE}`

I hadn't actually seen the Maven-style approach anywhere in YAML, but as you can see below it (or at least similar variants) does exist. It's kind of hard to search for this sort of thing on Github, unfortunately. Here's a quick Google survey (note that these capture lots of false positives from text and stuff around the code on these platforms):

image

(Following seems to catch `${env.`, `${env,`, `${env:`, etc)
image

image

esjewett:
BTW, don't get me wrong. It would be very cool if UI5-tooling just chose one and supported it out of the box to populate `process.env`, possibly even with a default environment variable name based on the task/middleware name and the configuration variable. Then if a particular plugin wanted to override `process.env`, it could, of course.

petermuessig:
@esjewett - I would be also fine to support this syntax. In the Maven world the usage of the prefix `env.` was a clear separation between accessing the env variables and the Maven properties. So I could also foresee that with the `${XXX}` you might also access other properties from the yaml maybe in a central properties location and use `${env.XXX}` to refer to the environment variables. IMO this would be more future proof.

PR #180, opened 2020-04-16
Support for environment variables and .env for baseUri and strictSSL

esjewett:
This pull request provides support in ui5-middleware-simpleproxy for specifying the `baseUri` and `strictSSL` configuration via environment variable or `.env` file. This addresses issue #174 for this package.

Also:
- Removes repetitive existence-checked access of `options.configuration`
- Adds a existence check on access of `options.configuration.debug` that is now required because it is possible to run without a configuration section

### DIFF
--- a/packages/ui5-middleware-simpleproxy/lib/proxy.js
+++ b/packages/ui5-middleware-simpleproxy/lib/proxy.js
@@ -61,8 +61,6 @@ module.exports = function ({ resources, options }) {
     options.configuration ? options.configuration.strictSSL : undefined
   );
 
-  log.info(providedStrictSSL);
-
   /*
   return function (req, res, next) {
       // [...]
--- a/packages/ui5-middleware-simpleproxy/lib/proxy.js
+++ b/packages/ui5-middleware-simpleproxy/lib/proxy.js
@@ -6,7 +6,35 @@ dotenv.config();
 
 const env = {
   baseUri: process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_BASEURI,
-  strictSSL: !(process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL === "false")
+  strictSSL: process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL
+}
+
+/**
+ * Handle decision between configuration and environment value while processing
+ * string values from environment and possible null and undefined values. Any string
+ * passed to environment variable except `"false"` will default to `true`. If both
+ * values are undefined or null, return `true` as well.
+ *
+ * @param {?boolean} environmentValue Value of the environment variable 
+ *                                  UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL
+ * @param {?boolean} configurationValue Value from the ui5.yaml configuration
+ * @returns {boolean} Indicator whether to require strict SSL checking
+ */
+function deriveStrictSSL(environmentValue, configurationValue) {
+  const environmentBooleanOrNull =
+    (environmentValue === undefined || environmentValue === null)
+      ? undefined
+      : !(environmentValue === "false");
+
+  if (environmentBooleanOrNull === undefined) {
+    if (configurationValue === undefined || configurationValue === null) {
+      return true;
+    } else {
+      return configurationValue;
+    }
+  } else {
+    return environmentBooleanOrNull;
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


petermuessig:
Hi @esjewett ,

did you see the change of @stockbal (#179) to use a mapping of the env variables in the `ui5.yaml`? Maybe this is something you also can reuse? It is more declarative to have such an env variable mapping rather than defining the env variables in the middleware. What do you think? 

BR,
Peter

esjewett:
I think it’s fine but I don’t think the syntax in that pull request is good so I would prefer to introduce an option like this without syntax until it can be better considered.

esjewett:
@vobu Will do today. Thanks 🙏.

petermuessig:
Just for my interest - what is your concern with respect to that API - it is aligned with the syntax existing in Maven. Do you expect another syntax closer related to the OS?

esjewett:
@petermuessig Ah, good to know that it comes from somewhere. It's been a decade since I used Maven. I would expect more of a shell-like syntax. It might be worth doing a quick survey of how projects do this in YAML config files before settling on a way you want to handle it. I think some projects do something like `!ENV ${MY_VARIABLE}`, but I don't really see why just `${MY_VARIABLE}` wouldn't work. That said, I also don't see a lot of advantage to it being configurable and some disadvantages like making it easier to make a mistake while setting it up.

esjewett:
@vobu README updated with variable names in the configuration section and a quick section on use of a `.env` file. I also added the `strictSSL` option to the configuration section.

packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -17,17 +25,20 @@ const log = require("@ui5/logger").getLogger("server:custommiddleware:proxy");
  * @returns {function} Middleware function to use
  */
 module.exports = function({resources, options}) {
+    const providedBaseUri = (options.configuration && options.configuration.baseUri) || env.baseUri;
+    const providedStrictSSL = (options.configuration && options.configuration.strictSSL) || env.strictSSL;
```
petermuessig, line 29: Hmm, this looks a bit tricky here - it works fine if only one of those values (either config or env) is defined - but in case of both values are defined, it behaves odd, assume:

`config.SSL = false || env.SSL = true` ==> env wins
`config.SSL = true || env.SSL = false` ==> config wins

What is the expected precedence behavior here? Config before env or env before config. I assume the last one should be the right one as the env vars should overrule the configs, right? 

By using the `${env.XXX}` syntax there would be a clear distinction on what is used. In this case here you need to do `undefined checks` and then fallback to the other value, e.g.:

Quick and dirty: 😄 
```js
// only consider strictSSL from env when a real value is set otherwise the value is undefined
const env = {
  baseUri: process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_BASEURI,
  strictSSL: (function(value) { return value !== undefined && !(value === "false")  })(process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL)
}
```
 
For the provided checks I would suggest to prefer the `env` variable and only of the env variable is undefined we fallback to the config, e.g.:

```js
const providedStrictSSL = env.strictSSL !== undefined ? env.strictSSL : (options.configuration && options.configuration.strictSSL);
```

What do you think?

petermuessig:
I am fine with your change even though I would prefer the mapping of the env variables as I suggested. But we can even add this later on - it can be added on top.

esjewett:
So, the clear precedence would be as follows (after discussion with Peter and forthcoming update to the pull request):

1. Value in an environment variable (or `.env`) specified in the YAML file (not implemented in this pull request)
2. Value in a default environment variable (or `.env`) specified in the library/documentation
3. Configuration value specified in the YAML file
4. Default value specified in the library/documentation

Additionally, if an environment variable value is populated, the value in the `.env` file should be ignored. This behavior is already handled by `dot-env`.

The ui5-tooling already handles 1. I think it should also handle 2 and arguably 3. This pull request implements 3 at the library (custom middleware) level.

Specifically (reversed in the latest update, but for posterity):

image

[Edit - discussed with Peter and updated the precedence above.]

packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -17,17 +25,30 @@ const log = require("@ui5/logger").getLogger("server:custommiddleware:proxy");
  * @returns {function} Middleware function to use
  */
 module.exports = function({resources, options}) {
+    // Environment wins over YAML configuration when loading settings
+    const providedBaseUri = env.baseUri || (options.configuration && options.configuration.baseUri);
+    // Handle fact that `strictSSL` is a boolean so we have to do existence checks
+    const falseyStrictSSL = 
+      (
+        env.strictSSL === undefined
```
petermuessig, line 33: `env.stricSSL` will never be `undefined` as `strictSSL: !(process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL === "false")` will be `true` in case of "`UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL=undefined`". That's why I made this function check in my suggestion.

esjewett: Yup, you're correct as I mentioned privately yesterday. I'm going to take the time to do this right, but probably can't until this weekend. I will withdraw this to draft status for the time being.

petermuessig: Hi Ethan, is there anything I can help you with? BR, Peter

esjewett: @petermuessig Just more time in the day. Can you help me with that? 😂 Hope to get this done today or tomorrow.

esjewett:
I've updated with a function including JSDoc to handle the logic around environment vs. configuration, string to boolean conversion, and `undefined` and `null` cases for both environment variables and configuration. In the case of configuration, this is not strictly necessary, but I figured I might as well be through.

I've also updated the `README.md` to make clear that environment will win over configuration.

I tested this pretty thoroughly, so hopefully we're good to go this time. I apologize for all the extra commits. This obviously needs to be squashed to get rid of all the thrashing around this issue.

packages/ui5-middleware-simpleproxy/lib/proxy.js
petermuessig: Looks good now. THX. :-)

PR #186, opened 2020-04-28
Ignore local routes

IObert:
Local routes break the middleware because the "webapp" folder doesn't exist.
The webapp is already served by the UI5 tooling itself, so this route isn't needed anylonger.


IObert:
Fixes SAP/generator-easy-ui5#32

PR #195, opened 2020-05-01
Automatically set a non-taken port for livereload server

resitamas:
Setting automatically an open port in case the `option.configuration.port` is not supplied. The free port selection is made by [portfinder](https://www.npmjs.com/package/portfinder), it scans the ports from _35729_ toward _65535_.


petermuessig:
Look good to me - I will merge and release it. Thanks for you contribution! :-)

PR #250, opened 2020-07-08
Enhance simpleproxy with adding configuration option for http headers

resitamas:
Add http headers from environment variable or from the configuration object

### DIFF
--- a/packages/ui5-middleware-simpleproxy/lib/proxy.js
+++ b/packages/ui5-middleware-simpleproxy/lib/proxy.js
@@ -102,7 +102,7 @@ module.exports = function ({ resources, options }) {
         proxyReqOpts.rejectUnauthorized = false;
       }
       if (providedHttpHeaders) {
-        proxyReqOpts.headers = Object.assign(proxyReqOpts.headers, providedHttpHeaders); 
+        Object.assign(proxyReqOpts.headers, providedHttpHeaders); 
       }
       return proxyReqOpts;
     },
--- a/packages/ui5-middleware-simpleproxy/lib/proxy.js
+++ b/packages/ui5-middleware-simpleproxy/lib/proxy.js
@@ -6,7 +6,8 @@ dotenv.config();
 
 const env = {
   baseUri: process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_BASEURI,
-  strictSSL: process.env.UI5_MIDDLEWARE_SIMPLE_PROXY_STRICT_SSL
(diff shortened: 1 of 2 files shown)


packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -37,6 +38,25 @@ function deriveStrictSSL(environmentValue, configurationValue) {
...
 
+/**
+ * Get the HTTP headers from environment variable if exists, otherwise get from the configuration
+ * 
+ * @param {string} environmentValue The value coming from the enviroment variable 'UI5_MIDDLEWARE_HTTP_HEADERS'
+ * @param {Object} configuration The configuration object
+ * 
+ * @returns {Object} http headers
+ */
+function getHttpHeaders(environmentValue, configuration) {
```
petermuessig, line 49: I have only a very little comment here => the first parameter is the http headers value of the environment and the second parameter is the configuration object. I would have either provided the environment object and the configuration object or the http header value from the env and the http header value from the config. But this is really reviewing on a nitpicking level from me now... :-)

packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -82,6 +101,9 @@ module.exports = function ({ resources, options }) {
       if (providedStrictSSL === false) {
         proxyReqOpts.rejectUnauthorized = false;
       }
+      if (providedHttpHeaders) {
+        proxyReqOpts.headers = Object.assign(proxyReqOpts.headers, providedHttpHeaders);
```
petermuessig, line 105: In case of using `Object.assign` the first parameter is the target which will be extended with the values from the additional parameters. In this case here it doesn't make a difference as you assign the return value of the `Object.assign` to the first parameter. Just FYI if you want to make a copy the first parameter of `Object.assign` should be an empty object.

packages/ui5-middleware-simpleproxy/lib/proxy.js
resitamas: My idea was that only pass the minimal values which are required in the function. Regarding the environment variables the others are not needed, but for the configuration object the _debug_ property is used in the function as well. I could have pass this as a third parameter, but I don't like to add boolean values as function parameters.

packages/ui5-middleware-simpleproxy/lib/proxy.js
resitamas: I know you said I don't need to do changes, but I removed this returning 😁

petermuessig:
Pushed to npm with:
 - ui5-middleware-simpleproxy => 0.4.0

PR #253, opened 2020-07-14
fix: default to debug false

jkoenig134:
Today I updated the livereload middleware for my project and the livereload-server was logging verbosely although I disabled that via configuration.
It took some debugging for me to detect the issue. 
`if (options.configuration && options.configuration.debug) {` is only true if debug is configured as true. This causes that the configured `false` will never be written in let debug. Simple workaround for that is defaulting the debug variable to false (I anyway don´t see the point for enabling it).


vobu:
🙌 thx for your contribution!

PR #311, opened 2020-09-23
fix(livereload): ensure proper concat for extra exts

petermuessig:


petermuessig:
The fix is not really necessary I noticed, but it improves the logging in the console which looks like that right now:

```
LiveReload is waiting for a browser to connect...

Protocol version: 7
Exclusions: /\.git\//,/\.svn\//,/\.hg\//
Extensions: xml,json,propertieshtml,css,js,png,gif,jpg,php,php5,py,rb,erb,coffee
Polling: false
```

and with the fix is looks like this:

```
LiveReload is waiting for a browser to connect...

Protocol version: 7
Exclusions: /\.git\//,/\.svn\//,/\.hg\//
Extensions: xml,json,properties,html,css,js,png,gif,jpg,php,php5,py,rb,erb,coffee
Polling: false
```

I thought it may also have a functional impact but luckily it is just the logging.

vobu (approved):
LGTM - you got that eagle eye for details :)

PR #338, opened 2020-11-12
feat: testing galore

vobu:
this PR introduces the full perspective of available testing frameworks to the showcase.

navigation-, interaction- and binding-tests against the app are coded as similar as possible in
- OPA5
- UIveri5
- wdi5

also, utilizing the tests and frameworks in a CI environment is demonstrated via respective tasks (`test:opa5-ci`,`test:ci`) and their use in github actions.

a blog post describing the setup along with a comparison of the frameworks with their pros and cons is in the making 😄


package.json
```diff
@@ -15,8 +15,15 @@
     "clean": "cd packages/ui5-app && yarn clean",
```
petermuessig, line 15: little hint: once your PR is in, I will switch here to the yarn workspace syntax, e.g.:

`yarn workspace ui5-app clean`

packages/ui5-app/package.json
```diff
@@ -32,12 +33,20 @@
   },
   "devDependencies": {
     "@openui5/ts-types": "^1.60.5",
```
petermuessig, line 35: `@openui5/ts-types` and `@sapui5/ts-types` - only one should be in - I would suppose we use the latest @sapui5 one

packages/ui5-app/ui5-min.yaml
```diff
@@ -0,0 +1,56 @@
+specVersion: "2.0"
+metadata:
+  name: ui5-app
+type: application
+framework:
+  name: OpenUI5
+  version: "1.77.1"
```
petermuessig, line 7: hmm, maybe we synchronize this with the types? 1.82?

packages/ui5-app/ui5-min.yaml
```diff
@@ -0,0 +1,56 @@
...
+      removeETag: true
+  - name: ui5-middleware-cfdestination
+    afterMiddleware: compression
+    configuration:
+      debug: true
+      port: 1091
+      xsappJson: "xs-app.json"
+      destinations:
+      # check that the destination name (here: "backend") matches your router in xssppJson
+      - name: "backend"
```
petermuessig, line 51: isn't it duplicate then? In Karma an here? But the servers are running on different ports, right?

packages/ui5-app/webapp/test/e2e-wdi5/tests.test.js
```diff
@@ -0,0 +1,100 @@
...
+const list = {
+    selector: {
+        viewName: "test.Sample.view.Other",
+        id: "PeopleList"
+    }
+}
+
+const backButton = {
+    selector: {
+        id: "__page0-navButton"
```
petermuessig, line 29: Is this really a good practise to use generated ID's? `__page0` works for now but can be changed at any time also in the framework.

petermuessig (commented):
Basically ok, just some open points regarding the project configuration. The rest looks great!

packages/ui5-app/webapp/test/e2e-wdi5/tests.test.js
vobu: with a more recent version of wdi5, regex for ids are now possible :)

packages/ui5-app/ui5-min.yaml
vobu: unfortunately karma(-ui5) doesn't play together with an external proxy running on a different port (yet), so need to duplicate the external proxy in karma conf itself.
see also: https://github.com/SAP/karma-ui5/issues/125

packages/ui5-app/package.json
vobu: done

package.json
vobu: roger that

packages/ui5-app/ui5-min.yaml
petermuessig: OK, thanks

petermuessig (commented):
Did you miss to upload your change?

vobu:
> Did you miss to upload your change?

still on it :)

petermuessig (approved):

Issue #359, opened 2020-11-24
Extending ui5-middleware-livetranspile to support different file types (e.g. TS support)


texttechne:
Hi,

I would like to propose a change to make this middleware more generic. Currently it is restricted to consume `.js` files and I do at least require the consumption of TS files. I think this is the last step to make the middleware a full-blown transpiler enabling all Babel features.

The change itself: Switch from static lookup ([`resource.byPath`](https://github.com/petermuessig/ui5-ecosystem-showcase/blob/master/packages/ui5-middleware-livetranspile/lib/livetranspile.js#L69)) to configurable globbing pattern `.{js,ts}`.

The configuration I would imagine in `ui5.yaml` along these lines: `pattern: ".{js,ts}"` (default = ".js")

Of course, there will be a performance penalty for the dynamic lookup. I'd assume it would be neglible. We could also switch between globbing and static lookup depending on the pattern: either ".js" / ".ts" or ".{js,ts}"...

Another question: do you prefer to define / import all Babel dependencies in the middleware? That's currently more or less the case. I don't see any real problem with this approach and it makes the DX more comfortable. However, it feels kind of wrong... But it would allow to put the essential and most useful Babel transforms into the middleware and enable it by features flag via `ui5.yaml`. So is this the way to go or do you would like to create a separate livetranspile-middleware?

I would put together the PR if you're fine with this feature and you outline the direction to take :-)

vobu:
hi, all of your suggestion sound good and are fine-by-me™ 😆 
and definitely please work on the existing `livetranspile` middleware, don't add more animals to the zoo!
couple of remarks:
- configureable file extension with `js` default is fine, but honestly, what other than `.{js,ts}` is out there? 
  -> why not just hard-wire it to `js|ts`?
- the point about the babel dependencies i'm not really getting: do you mean the dependencies in `package.json` (à la https://github.com/petermuessig/ui5-ecosystem-showcase/blob/d4eddf7fc1b64f832da5059d3d414f8ae454f1c7/packages/ui5-middleware-livetranspile/package.json#L12) or more the default babel config option (à la https://github.com/petermuessig/ui5-ecosystem-showcase/blob/d4eddf7fc1b64f832da5059d3d414f8ae454f1c7/packages/ui5-middleware-livetranspile/lib/livetranspile.js#L41)?
best, v.

htammen:
[...]

PR #364, opened 2020-11-25
feat: allow for file patterns as JS input source

texttechne:
This PR affects ui5-middleware-livetranspile as well as ui5-task-transpile.
It allows for Babel consumption of different file types other than JS.
See #359 

By default only JS files are considered. The pattern for file lookup can be configured in `ui5.yaml` via attribute `filePattern`, e.g.:  ```filePattern: ".{js,ts}"```

### Caveat
ui5-task-transpile will emit the source files for transpilation, e.g. Component.ts, in addition to the transpiled JS files.
Can be worked around by adding rimraf to the build script in package.json:

```
"scripts": {
    "serve": "ui5 serve",
    "build": "ui5 build && rimraf dist/**/*.ts"
  }
```
Are there better ways to make this happen by means of the UI5 tooling API?

### DIFF
--- a/packages/ui5-middleware-livetranspile/lib/livetranspile.js
+++ b/packages/ui5-middleware-livetranspile/lib/livetranspile.js
@@ -24,60 +24,71 @@ fileNotFoundError.file = ""
  * @returns {function} Middleware function to use
  */
 module.exports = function ({ resources, options }) {
-    const plugins =
-        options.configuration && options.configuration.transpileAsync
-            ? [
+    const config = options.configuration || {}
+    const plugins = config.transpileAsync
+        ? [
+              [
+                  "babel-plugin-transform-async-to-promises",
+                  {
+                      inlineHelpers: true
+                  }
+              ]
+          ]
+        : []
+    const babelConfig = config.babelConfig
+        ? config.babelConfig
+        : {
+              sourceMaps: "inline",
+              plugins,
+              presets: [
                   [
-                      "babel-plugin-transform-async-to-promises",
+                      "@babel/preset-env",
                       {
-                          inlineHelpers: true
+                          targets: {
+                              browsers: "last 2 versions, ie 10-11"
+                          }
                       }
                   ]
               ]
-            : []
-    const babelConfig =
-        options.configuration && options.configuration.babelConfig
-            ? options.configuration.babelConfig
-            : {
-                  sourceMaps: "inline",
-                  plugins,
-                  presets: [
-                      [
-                          "@babel/preset-env",
-                          {
-                              targets: {
-                                  browsers: "last 2 versions, ie 10-11"
-                              }
-                          }
-                      ]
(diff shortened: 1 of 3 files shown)


packages/ui5-middleware-livetranspile/lib/livetranspile.js
```diff
@@ -24,60 +24,71 @@ fileNotFoundError.file = ""
...
-                      ]
-                  ]
-              }
-
+          }
+    const filePatternConfig = config.filePattern || ".js"
     return (req, res, next) => {
         if (
             req.path &&
             req.path.endsWith(".js") &&
```
vobu, line 58: shouldn't this also be sth like `filePatternConfig.forEach(ending => req.path.endsWith(ending))` in order to act on all configured file endings? haven't validated, just by looking at the code.

packages/ui5-middleware-livetranspile/lib/livetranspile.js
```diff
@@ -24,60 +24,71 @@ fileNotFoundError.file = ""
...
             !req.path.includes("resources/") &&
-            !((options.configuration && options.configuration.excludePatterns) || []).some((pattern) =>
-                req.path.includes(pattern)
-            )
+            !(config.excludePatterns || []).some((pattern) => req.path.includes(pattern))
         ) {
             const pathname = parseurl(req).pathname
+            const pathWithPattern = pathname.replace(".js", filePatternConfig)
```
vobu, line 63: probably not necessary if above comment holds true

packages/ui5-middleware-livetranspile/lib/livetranspile.js
```diff
@@ -24,60 +24,71 @@ fileNotFoundError.file = ""
  * @returns {function} Middleware function to use
  */
 module.exports = function ({ resources, options }) {
-    const plugins =
-        options.configuration && options.configuration.transpileAsync
-            ? [
+    const config = options.configuration || {}
```
vobu, line 27: thanks for that!

vobu (commented):
just a minor detail

packages/ui5-middleware-livetranspile/lib/livetranspile.js
texttechne: Actually it is not necessary, because req.path comes from the frontend calls to the node.js server. And the frontend only knows about js files. I cannot imagine a use case where the frontend wants to get the non-transpiled files directly... maybe just a lack of imagination, but a different feature request for sure :grinning:

packages/ui5-middleware-livetranspile/lib/livetranspile.js
texttechne: Thats the actual place where the pattern is needed: for the file lookup of the requested js source

vobu (approved):
thanks for the explanations

PR #371, opened 2020-11-27
feature: mvn-dependency-provider

Siolto:
This custom task downloads ui5 sources (think: libraries) masked as maven .jar dependencies and unpacks them to a location of your choice.


vobu (commented):
please @petermuessig also have a look - this is a use-case applying also to SAP internally.
Personally I'm fine with the  optional chaining op `?` as we have Node.js LTS 14.x in `.nvmrc` - but want to get your input here as well.

petermuessig:
May to start with a question: what is finally the purpose of this task? Do you want to consume Maven dependencies in the build, for the server? That would be interesting to understand first.

Because I also thought about that topic in the past. My idea is a bit different. Instead of spending a task for this I would see Maven as an orchestrator around which uses Maven plugins to place and eventually unpack the JARs in a special directory and then creates the dependencies.yaml for the project which finally allows to use those resources in the build and serve seamless like NPM packages. Then the Maven plugin runs Node. In your case it is vice-versa. You are calling Maven to unpack the dependencies but then this becomes a task in the middle of the build.

vobu:
as far as i understood it, the tasks's purpose is to include .jar-packed ui5 libraries at build-time for serving them at runtime - essentially a bridge for using ui5 libs similar to npm mechanisms, but w/o npm. 
so the workflow is essentially
ui5 build ... --include-task=theMvnTask
(unpacks to somewhere in /webapp)
ui5 serve

petermuessig:
Why not then vice-versa to run a Maven command which unpacks the dependencies and then run `ui5 serve` either within Maven (one option is the eirslett plugin or I can try to OSS my node-maven-plugin in future) or as separate tasks? Running mvn could also be one of the npm scripts maybe?

petermuessig:
Please don't misunderstand me - I am not against it - I just want to discuss if it is the right place to put this.

Siolto:
Valid points @petermuessig. We decided to publish the custom UI5 task separately see [here](https://github.com/js-soft/ui5-task-mvn-dependency-provider). I will close this pull request.

Issue #372, opened 2020-11-27
Consistent Code Formatting of Source Code


texttechne:
Hi,

having written the first PR, I must confess, that I've had problems to conform to the different code styles of the different packages.

Hence I would like to propose the following changes:
* decide on one code style 😄 (by the way, I prefer semi-colons at the end of the statement 😉)
* install & configure prettier & eslint in root package
* reformat all packages and commit (of course, this interfers with a clean git history, but better now than never)
* install lint-staged / husky to enforce the formatting & linting rules on every commit
  - https://github.com/okonet/lint-staged

Most likely, you know about all this & haven't had the time, so think of this issue as a friendly reminder 😄

vobu:
*me opens the gate* 
let the trolls roll in ;)

petermuessig:
OMG, you said the bad words, @texttechne  => **semicolons**! => 🏃 => @vobu will find you ... 🤣 

Till now, we loved the anarchy in our repository and we believe in being the better formatters... 😄 But recently also @vobu added `prettier` and with that we have maybe started the discussions - the trolls are already in... 

I am open to discuss - but honestly, I am not a real fan of it... Maybe I just need some time to look into them and play around with them so they already do what I want and not try to correct me always. It is nothing worse than a prettier which adopts my lovely formatted code... 🤣

vobu:
hell yeah, i _will_ find you 🤣✌️

texttechne:
da habe ich ja ein Faß aufgemacht 🤣 

Ok, I should have asked before assuming. I did see chaos, but actually only package `ui5-task-transpile` is in some disarray. Apart from this package all others adhere to their own style definition. The good news: it essentially boils down to the following basic style definitions:
```
{
  "printWidth": 120,                     <-- the same for every package
  "tabWidth": 4,                            <-- 2 or 4
  "semi": false,                               <-- either false @vobu or just leave it out (default = true) 
  "trailingComma": "none"           <-- either "none" or "es5" or leave it out (default = es5)
}
```

So instead of defining a `.prettierrc` on the root level, it must be defined in each package separately. Then you maintain the freedom / anarchy you already live 😄 

I've prepared a PR for this setup, so that you can toy around with it.

### Arguing for Prettier
To be honest: I'm opinionated about code formatting. So let me argue for the use of prettier just this once. Afterwards I shut up and troll my way...

The best thing about prettier: it's opinionated! 😁 
[...]

PR #373, opened 2020-11-28
feature(Build): Add Prettier & editorconfig

texttechne:
Proposal to integrate prettier & editorconfig.
Each package defines its own `.prettierrc` to support different styles.

Based on discussion of #372 

Don't merge, this PR is just a proposal and needs some more love to fine-tune the details.

### Noteworthy Stuff
The root package defines `.editorconfig` to define the Basics like usage of utf-8. 
It can also be used to define indentation for file types in all modules, e.g. package.json or html files, ...

The root package also encompasses a `.prettierignore` file. I chose to ignore the complete ui5-app module, because of the templates within it. It would need some configuration & testing to get prettier to work here; then again, prettier could well be the wrong choice for templates.

### To Toy Around
* Add the following plugins to your IDE:
   * editorconfig
   * prettier
* Go to any module and format a JS file
* Try out "format on save" option.
* Use `// prettier-ignore` to disable prettier for the next statement.

### TODO
* Most `package.json` files are formatted differently to their actual definition
  * set indentation globally via editorconfig?
  * use "overrides" in each prettierrc file?
* Most YAML files are using single quotes
  * use "overrides" in each prettierrc file?
  * live with double quotes?
* probably more...


texttechne:
Git history was messed up => close

PR #374, opened 2020-11-28
feat(Build): integrate prettier & editorconfig

texttechne:
Proposal to integrate prettier & editorconfig.
Each package defines its own .prettierrc to support different styles.

Based on discussion of #372

Don't merge, this PR is just a proposal and needs some more love to fine-tune the details.

## Noteworthy Stuff

The root package defines .editorconfig to define the Basics like usage of utf-8.

The root package also encompasses a .prettierignore file. As an example I chose to ignore the complete ui5-app module, because of the templates within it. It would need some configuration & testing to get prettier to work here; then again, prettier could well be the wrong choice for templates.

### To Toy Around

* Add the following plugins to your IDE:
   - editorconfig
   - prettier
 * Go to any module and format a JS file
 * Try out "format on save" option.
 * Use // prettier-ignore to disable prettier for the next statement.

### TODO
* Base files like package.json, .prettierrc
* YAML in general: maybe it's best to ignore yaml completely
* probably more...


petermuessig:
`.editorconfig`is not integrated with #484  - next stop `prettier`

PR #412, opened 2020-12-15
feat(livereload): add exclusions option to livereload middleware

dominikfeininger:
- multiple exclusions
- single exclusion

can be passed through to liveserver module


petermuessig (approved):
LGTM - thanks Dominik

PR #466, opened 2021-01-26
chore: dependency version changes only

petermuessig:


vobu:
abadoning this in favor of the dependabot PRs

PR #484, opened 2021-02-07
chore: use .editorconfig

petermuessig:


petermuessig:
@texttechne - here this first part of your suggestion comes

vobu (approved):

PR #485, opened 2021-02-09
fix: non-specified exclusions crash the server

vobu:
if no configuration at all is specified, the middleware used to crash `ui5 serve`.
(we'll optional-chain such cases once node 14 lts is more widely adopted)

### DIFF
--- a/packages/ui5-middleware-livereload/lib/livereload.js
+++ b/packages/ui5-middleware-livereload/lib/livereload.js
@@ -46,7 +46,10 @@ module.exports = async ({ resources, options }) => {
         watchPath = options.configuration.watchPath || options.configuration.path;
     }
     let exclusions = [];
-    const aOptExclusions = options.configuration.exclusions;
+    let aOptExclusions;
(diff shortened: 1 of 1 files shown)


packages/ui5-middleware-livereload/lib/livereload.js
```diff
@@ -46,7 +46,10 @@ module.exports = async ({ resources, options }) => {
         watchPath = options.configuration.watchPath || options.configuration.path;
     }
     let exclusions = [];
-    const aOptExclusions = options.configuration.exclusions;
+    let aOptExclusions;
+    if (options.configuration && options.configuration.exclusions) {
+        aOptExclusions = options.configuration.exclusions;
+    }
     if (options.configuration && aOptExclusions && Array.isArray(aOptExclusions)) {
```
petermuessig, line 53: you could have removed the options.configurations check then in the aOptExclusions check - but anyway - this middleware may need a bit of refactoring soon. I will release it like this for now.

petermuessig (approved):

PR #492, opened 2021-02-15
Add auth to routes

d-sooter:
Add auth support. 

The user can pass the authenticationMethod via the options 
and the authentication method defined in the routes is respected and used

### DIFF
--- a/packages/ui5-middleware-cfdestination/README.md
+++ b/packages/ui5-middleware-cfdestination/README.md
@@ -53,6 +53,7 @@ server:
       afterMiddleware: compression
       configuration:
         authenticationMethod: none # none or route default is none
+        allowServices: false # allows sap services like SAP iot to be used 
         debug: true
         port: 1091
         xsappJson: "xs-app.json"
--- a/packages/ui5-middleware-cfdestination/README.md
+++ b/packages/ui5-middleware-cfdestination/README.md
@@ -52,8 +52,8 @@ server:
     - name: ui5-middleware-cfdestination
       afterMiddleware: compression
       configuration:
-        authenticationMethod: none # none or route default is none
-        allowServices: false # allows sap services like SAP iot to be used 
+        authenticationMethod: "none" # "none" || "route", default: "none"
+        allowServices: false # allows sap services like SAP IoT to be used 
         debug: true
         port: 1091
         xsappJson: "xs-app.json"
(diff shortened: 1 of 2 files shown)


packages/ui5-middleware-cfdestination/lib/cfdestination.js
```diff
@@ -25,16 +25,15 @@ module.exports = function ({ resources, options }) {
     const xsappConfig = JSON.parse(fs.readFileSync(options.configuration.xsappJson, 'utf8'));
     const xsappPath = options.configuration.xsappJson.replace("xs-app.json", "");
 
-    // the embedded approuter (the one that runs locally) will ignore all auth mechanisms. 
-    // Even when xsappConfig.authenticationMethod = "route" and only define the routes, 
-    // the approuter will try to read the env var during startup - and fail if they are missing
-    xsappConfig.authenticationMethod = "none";
+    // the default auth mechanism is set to none but the user can pass an auth method using the options
+    xsappConfig.authenticationMethod = options.configuration.authenticationMethod || "none";
```
vobu, line 29: you need to do the optional dance here in order not to have the module crash at runtime:
`(options.configuration && options.configuration.authenticationMethod) || "none"`

packages/ui5-middleware-cfdestination/README.md
```diff
@@ -52,6 +52,7 @@ server:
     - name: ui5-middleware-cfdestination
       afterMiddleware: compression
       configuration:
+        authenticationMethod: none # none or route default is none
```
vobu, line 55: comma + quotes missing (pedantry, i know 😛 )
`# "none" || "route", default: "none"`

vobu (changes requested):
only small change necessary

packages/ui5-middleware-cfdestination/lib/cfdestination.js
d-sooter: Is that really needed? As far as i can tell the configuration object is always there. If not it would already crash on line 22. 
I can add it but i dont see that its needed.

packages/ui5-middleware-cfdestination/README.md
d-sooter: Done

packages/ui5-middleware-cfdestination/lib/cfdestination.js
vobu: right on, point taken

vobu (approved):

PR #516, opened 2021-04-22
missing encoding of uri parameters

sebbi08:
When adding uri parameters to the proxy request the "req.query" parameter object contains the parameters unescaped and the finaly build uri of them will also be unescaped


vobu (approved):
LGTM!

PR #527, opened 2021-06-06
feat: add default options for cfdestination

vobu:
this PR adds a set of default options to `ui5-middlware-cfdestination`, so that an empty config of the custom middleware can be put into `ui5.yaml`:
```yaml
server:
  customMiddleware:
      afterMiddleware: compression
```


vobu:
pushing this through to have it available for the upcoming blog post

PR #546, opened 2021-11-12
sap-client shall be configured with quotes

rygiol:
If the sap-client number starts with zero (e.g. 001) and it got configured here without quotes, then the proxy will try to connect to sap-client 1 instead of 001.
Furthermore, the proxy log will not point the developer to this wrong sap-client 1. It is quite hard to figure out the wrong configuration.
Hence, we shall give a hint in this documentation, that the client shall be entered always with quotes.

### DIFF
--- a/packages/ui5-middleware-simpleproxy/README.md
+++ b/packages/ui5-middleware-simpleproxy/README.md
@@ -69,7 +69,7 @@ server:
       httpHeaders:
         Any-Header: AnyHeader
       query:
-        sap-client: 206
+        sap-client: '206'
       excludePatterns:
       - "/local/**"
 ```


petermuessig:
Hi @rygiol , that`s true when there are no quotes used the sap-client is intepret as number and thus leading zeros will be removed. Good hint, contribution accepted! Thanks so much, Peter

petermuessig:
Will need to check the errors in the build first.

rygiol:
@petermuessig: The pipeline seems to an have with the long commit message. I will drop this PR and will create another commit.

petermuessig:
It's not only your commit message, there is something wrong with the build which is even not related to your change. But fixing the commit message isn't bad at all... Thx... :-)

rygiol:
I've just did a direct change in the main branch... not sure, why the PR process was not triggered for me this time. I guess I am part of the contributors now. Anyway, since the change is not affecting any productive code, I hope such a readme change is fine.

petermuessig:
Thanks for the contribution, I will check for the build results.

PR #559, opened 2022-01-19
Fixed error in ui5-tooling-modules

L4Ndoo:
Fixed error in ui5-tooling-modules\lib\util.js which made the build task fail completly when using ui5-tooling-modules


vobu (approved):

vobu:
even though the `commitlint` fails, I'll go ahead and merge this as it's a logging/stdout enhancement
→ nice error messages FTW!
thanks for spotting that detail @L4Ndoo 🙌

PR #572, opened 2022-02-08
Fix copy paste error

sebbi08:


petermuessig:
Good catch! Thx @sebbi08

PR #573, opened 2022-02-09
refactor(component): added .env file to examples of onelogin

uxkjaer:
Added url to handle the login to the SAPES5 system for the examples


petermuessig (changes requested):
Adding the URL is fine, but do you really want to submit your username and password? If this was by mistake, please update with amend and force push to remove it from git history...

PR #579, opened 2022-02-22
feat(SimpleProxy):  added a feature to skip the server cache 

uxkjaer:
The skipcache flag is introduced to remove the ~<GUID>~ cache when serving the embedded FLP
launchpad and it matches an exclude pattern. Also fixed a bug on the excludepattern when using an
array with more than one entry as it was always returning true.

### DIFF
--- a/packages/ui5-middleware-simpleproxy/lib/proxy.js
+++ b/packages/ui5-middleware-simpleproxy/lib/proxy.js
@@ -145,10 +145,17 @@ module.exports = function ({ resources, options }) {
     https: protocol === "https",
     limit: limit,
     filter: excludePatterns && Array.isArray(excludePatterns) && function(req, res) {
-      return excludePatterns.some(glob => {
+      const pattern = !excludePatterns.some(glob => {
         isDebug && log.info(`Proxy request ${req.url} is matched to glob "${glob}": ${minimatch(req.url, glob)} ${providedBaseUri}`);
-        return !minimatch(req.url, glob);
+
+        return minimatch(req.url, glob);
       });
+      if (!pattern && req.url.includes("~") && (options.configuration && options.configuration.skipCache)) {
+        const newUrl = req.url.replaceAll(/\/~.*~.\//g, "/");
+        isDebug && log.info(`Removing cache from ${req.url}, resolving to ${newUrl}`);
+        req.url = newUrl;
+      }
+      return pattern
     },
     preserveHostHdr: false,
     proxyReqOptDecorator: function (proxyReqOpts) {
--- a/packages/ui5-middleware-simpleproxy/README.md
+++ b/packages/ui5-middleware-simpleproxy/README.md
@@ -30,6 +30,8 @@ npm install ui5-middleware-simpleproxy --save-dev
   Query parameters set for the proxied request. Will overwrite the parameters from the request. 
 - excludePatterns: `string[]`
   Array of exclude patterns using glob syntax
+- skipCache: `boolean`
+  Remove the cache guid when serving from the FLP launchpad if it matches an excludePattern
 
 In general, use of environment variables or values set in a `.env` file will override configuration values in the `ui5.yaml`.
 


packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -145,10 +145,17 @@ module.exports = function ({ resources, options }) {
...
     limit: limit,
     filter: excludePatterns && Array.isArray(excludePatterns) && function(req, res) {
-      return excludePatterns.some(glob => {
+      const pattern = !excludePatterns.some(glob => {
         isDebug && log.info(`Proxy request ${req.url} is matched to glob "${glob}": ${minimatch(req.url, glob)} ${providedBaseUri}`);
-        return !minimatch(req.url, glob);
+
+        return minimatch(req.url, glob);
       });
+      if (!pattern && req.url.includes("~") && (options.configuration && options.configuration.skipCache)) {
```
petermuessig, line 153: why not just using the following command: 

`const newUrl = req.url.replaceAll(/\/~.*~\//g, "/");`

packages/ui5-middleware-simpleproxy/lib/proxy.js
```diff
@@ -145,10 +145,17 @@ module.exports = function ({ resources, options }) {
...
+        const newUrl = req.url.slice(0,req.url.indexOf("~")) + req.url.slice(req.url.lastIndexOf("~") + 3,req.url.length)
+        isDebug && log.info(`Removing cache from ${req.url}, resolving to ${newUrl}`);
+        req.url = newUrl
+      }
+      return pattern
```
petermuessig, line 158: I need to add prettier to enforce code conventions (Semicolon or not semicolon => that's the question!)... 😜

petermuessig (changes requested):
I would prefer the regex instead of using indexOf to remove the cachebuster tokens

packages/ui5-middleware-simpleproxy/lib/proxy.js
uxkjaer: I've added semicolons as I could see that was the convention. I prefer not to have them though ;)

If i did a prettier, it would make changes to the entire file. Maybe add lint-staged and a seperate package with the prettier config?

petermuessig: 👍 - we could do that in a separate change - in general, I think using prettier in one or the other direction may be the best option and enforce it with a commit hook or so to come to "one" style when many people are working together on one project... If you are OK with going for it later then we submit this one and do the prettier stuff independently?

petermuessig (approved):

packages/ui5-middleware-simpleproxy/lib/proxy.js
uxkjaer: yeah sounds good. I'll open an issue. We can probably get many opinions on the prettier config as well :crossed_swords:

uxkjaer:
will you do the version bump for pubslishing or is this handled via CI?

petermuessig:
The release process is explained here:
https://github.com/ui5-community/ui5-ecosystem-showcase/blob/main/CONTRIBUTING.md#release-process

You can try it out or I can do it a bit later today...

uxkjaer:
I'm heading to bed now. I'd appreciate if you did it. I assume I'll also need to be able to push changes to the repo to do it right?

petermuessig:
Included here:

```
lerna success published 13 packages
 - ui5-middleware-cfdestination@0.8.2
 - ui5-middleware-livecompileless@0.2.3
 - ui5-middleware-livetranspile@0.3.6
 - ui5-middleware-onelogin@1.1.0
 - ui5-middleware-servestatic@0.3.9
 - ui5-middleware-simpleproxy@0.8.0
 - ui5-middleware-stringreplacer@0.3.1
 - ui5-task-compileless@0.2.4
 - ui5-task-pwa-enabler@0.2.8
 - ui5-task-stringreplacer@0.6.1
 - ui5-task-transpile@0.3.6
 - ui5-task-zipper@0.4.8
 - ui5-tooling-modules@0.1.3
```

Patch set follows due to dependencies issues with the yarn.lock.

petermuessig:
The following tasks and middlewares require updates due to previous issue in yarn.lock:

```
 - ui5-middleware-webjars: 0.3.4 => 0.3.5
 - ui5-task-zipper: 0.4.8 => 0.4.9
```

petermuessig:
All done:

```
Successfully published:
 - ui5-middleware-webjars@0.3.5
 - ui5-task-zipper@0.4.9
```

PR #583, opened 2022-02-23
feat(ui5-tooling-modules): ignore UI5 modules for transpiling

petermuessig:
Enhanced the transpiling script to also follow the dependencies
chain, e.g. if a module is resolve from another NPM package (not
via the UI5 tooling) then the dependencies of this UI5 module needs
to be also covered by the task. For the middleware the client will
follow the dependencies and request them from the server.

### DIFF
--- a/packages/ui5-app-module/Chart.js
+++ b/packages/ui5-app-module/Chart.js
@@ -1,0 +1,63 @@
+sap.ui.define(["sap/ui/core/Control", "chart.js/auto"], function(Control, Chart) {
+
+    return Control.extend("ui5-app-module.Chart", {
+        metadata: {
+            properties: {
+                "title": "string",
+                "color": "sap.ui.core.CSSColor"
+            },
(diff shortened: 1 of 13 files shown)


petermuessig:
@marcelschork @mauriciolauffer @marianfoo - check this out, I also added an NPM package providing UI5 controls as modules + a thirdparty library to display a chart in the sample app... 😉 

cc: @vobu

petermuessig:
Karma testing doesn't work yet as the UI5 resources are served from webapp which is not handled by the middlewares...

petermuessig:
Now also Karma works again, the reason was more the detection of the pathname via `req.path` which is not available via Karma middleware. Using now the `middlewareUtil.getPathname` to be on the safe side...

petermuessig:
BTW: the task of the tooling extension is now also able to resolve the dependencies from the XMLView!

PR #586, opened 2022-02-25
feat: use namespace instead of package name for reusable ui5 modules

marcelschork:
Hi @petermuessig,

really cool work regarding the ui5-tooling-modules 👍. Sry for anwsering a bit late, but I was very busy the last two days.  
Here is the other solution with namespaces, which is the one I prefer because it's more natural for a UI5 dev. Let's discuss pro's and con's with @marianfoo and @mauriciolauffer. 

Have a nice weekend :)
Marcel

### DIFF
--- a/packages/ui5-app-module/Chart.js
+++ b/packages/ui5-app-module/Chart.js
@@ -1,6 +1,6 @@
 sap.ui.define(["sap/ui/core/Control", "chart.js/auto"], function(Control, Chart) {
 
-    return Control.extend("ui5-app-module.Chart", {
+    return Control.extend("ui5.cc.chart.Chart", {
         metadata: {
             properties: {
                 "title": "string",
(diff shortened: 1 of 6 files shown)


petermuessig:
Hi @marcelschork,

should we create a second module aside using the shimming?

BR,
Peter

marcelschork:
Hi @petermuessig,
both of the options are valid and some devs probably prefer the namespace approach as this is what they know from UI5. So why not demonstrating both. Maybe we can have a (non control) ui5-module this time to showcase both type of artifacts. Shall I create a PR?

Have a nice and sunny weekend,
Marcel

petermuessig:
Hi @marcelschork 

yes, that's what I had in mind - maybe you can use the markdown control or any other to add it via this approach as custom control to the ecosystem showcase app.

THX and BR,
Peter

PR #591, opened 2022-03-22
Usage of replaceAll produces error while building

L4Ndoo:
Fixed the usage of replaceAll, which is not  available in Javascript, and switched to using replace, which also replaces all occurences of the provided regex since it is marked with a global modifier.


petermuessig:
Hi @L4Ndoo ,

good finding! Will merge it and release a new version. I'll update the commit message to follow the conventional commit rules.

THX and BR,
Peter

petermuessig:
Version 0.2.5 about to be released

PR #592, opened 2022-03-23
feat(ui5-middleware-servestatic): find-config dependency

rpanneel:
Hey @petermuessig / @vobu ,

This PR adds the find-config dependency to the middlewares ui5-middleware-servestatic and ui5-middleware-simpleproxy in order to add the possibility to move the .env-file to a parent directory.

This can be useful when creating a sort of app-host repository, which can contain multiple Fiori apps (via submodules or just plain old git clone). The app-host can then start the test-pages of the apps with only one .env configuration in the parent-folder.

When a .env-file is found in the app-folder it will use that one and don't look upward on the parent-tree. Let's discuss 😃 

Kind regards,

Robin


petermuessig:
Hi @rpanneel ,

thanks for your contribution. This is an interesting idea. I am wondering whether we should make this configurable to avoid the env lookup. Could there be side effects having a .env file in general. I had the same bad experience with the .yo-rc file of Yeoman to identify the root directory. I think the option to use it is great, but maybe better with a switch? 

WDYT? 

BR,
Peter

rpanneel:
> Hi @rpanneel ,
> 
> thanks for your contribution. This is an interesting idea. I am wondering whether we should make this configurable to avoid the env lookup. Could there be side effects having a .env file in general. I had the same bad experience with the .yo-rc file of Yeoman to identify the root directory. I think the option to use it is great, but maybe better with a switch?
> 
> WDYT?
> 
> BR, Peter

Hey @petermuessig,

I tested it locally and haven't noticed real side-effects, but I didn't test all the possible edge cases off course. Making it configurable might even be better, but the only downside in the request from our developers is the fact that they want to use the app-host (so the wrapper project) just to start the apps from a central place, the child apps shouldn't be modified. So when we change the config in the ui5.yaml we might have issues for the persons who clone the child project and run that. 

My initial idea was creating a script that would add the FLP sandbox config from all child apps and have one FLP Sandbox in the app-host parent project, but the problem is that app1 need to be tested with another test-user than app2 for example. This PR wouldn't solve this issue neither imo.

Maybe this request is really an edge case. At least it was the first time I got a request like that and at first I saw the benefit in implementing it, but how longer I think about it I have doubts...

WDYT?

Kind regards,

Robin

petermuessig:
Maybe we can do it the following way: use the recursive `find-config` by default and add an option to disable it to be on the safe side in case of errors occur.

WDYT?

rpanneel:
Good idea.
I will try to add it somewhere this week in my PR.

petermuessig:
Closing stale PR

PR #594, opened 2022-04-01
feat: include ui5-cc-md published on npm via it's namespace

marcelschork:
Hi @petermuessig,
like discussed in #586. I just used the custom control ui5-cc-md from @vobu to showcase how custom controls / modules can be included via it's namespace instead of package name + ui5-tooling-modules.


petermuessig:
Other examples are already incorporated - skipping this one here...

PR #600, opened 2022-04-08
chore: add metadata to `package.json` for all packages for ui5-community website

marianfoo:
To get a comprehensive metadata database as well as to display data in the UI5 community website, metadata is necessary in the respective packages.

For this purpose, a subitem is created in `package.json` with the name `ui5-community` similar to the already existing item `ui5`.

In this subitem two arrays are added with the name `types` and `tags`.
In `types` the types will be contained, which can be multiple for each package. For example: "task" or "middleware".
In `tags` several free descriptive names of the package can be entered.

The explicit specification of these parameters makes the extraction much easier, instead of pulling them out of yaml or json files.

### DIFF
--- a/packages/ui5-middleware-cfdestination/package.json
+++ b/packages/ui5-middleware-cfdestination/package.json
@@ -35,5 +35,13 @@
   },
   "ui5": {
     "dependencies": []
+  },
+  "ui5-community": {
+    "types": [
+      "middleware"
+    ],
+    "tags": [
+      "btp"
+    ]
   }
 }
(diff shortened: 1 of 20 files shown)


marianfoo:
@petermuessig 
As discussed:
Please check if the rough concept fits.
Then I would extend it to the other packages.
We still have to think about which tags we want to use.
There I am a little uncreative.

petermuessig (approved):
LGTM

petermuessig (approved):

PR #608, opened 2022-04-30
WIP: refactor: replace yarn with pnpm

uxkjaer:
replace yarn with pnpm as packagemanager and upgrade of all packages

BREAKING CHANGE: new package manager

	new file:   packages/ui3-middleware-onelogin/sample/example_embeddedflp/pnpm-lock.yaml


petermuessig:
Interesting reading: https://blog.logrocket.com/javascript-package-managers-compared/ - so maybe I would be also in favor to use `pnpm` - just reading the article! Storing dependencies only once on HDD is really interesting - getting rid of the node_modules trashbin in each project... 😁

PR #614, opened 2022-05-20
feat(ui5-tooling-modules): support name-spacing of modules

petermuessig:
When using the configuration option addToNamespace 3rd party modules
will now be moved into the project namespace in the thirdparty folders
and the used module names in JS and XML files will be prefixed with
the projectNamespace+thirdparty. This ensures that 3rd party modules
are packaged locally within the application and do not conflict globally
with other modules.


petermuessig:
@liwde - check this out

PR #617, opened 2022-05-27
fix(ui5-middleware/task-stringreplacer): handle missing strings

piejanssens:
If there were no strings provided, the build would fail after a while and the middelware would get in an endless loop and you'd have to kill the node ùi5 serve` process (and the live reload would keep running). 

With this PR a check is done and the user is warned in case of the middleware and there is no attempt to replace any strings. In the case of the build taks we raise an error and stop the build to avoid that builds are completed while missing strings intended for production.


petermuessig:
THX for the contribution @piejanssens, will squash and merge it.

petermuessig:
THX for the contribution @piejanssens, will squash and merge it.

PR #632, opened 2022-06-27
feat(ui5-tooling-modules): support app-local bundle defs

petermuessig:
To support special scenarios like `firebase` we need to create local bundle definitions to package all individual modules into a single bundle, otherwise this leads to problems that different modules aren't aware of each other. 

Now the ui5-tooling-modules can read project local bundle definitions via the app name + the path in the app folder, e.g. `ui5-app/lib/firebase`. With this module name the UI5 can also load the bundle via the `ui5-tooling-modules` extension.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -10,4 +10,5 @@
 /**/*.png
 
 # Additionally ignore folders for lint-staged
+/packages/ui5-app/lib/**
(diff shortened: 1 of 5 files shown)


marianfoo (approved):
LGTM!

PR #643, opened 2022-08-16
fix: exclude resources with OmitFromBuildResult tag from zip

hannaolb:
hi @petermuessig 
as we discussed this is the small change i propose to exclude files from the zip if the OmitFromBuildResult tag is set.


petermuessig (approved):
LGTM

PR #662, opened 2022-12-21
chore: update dependencies

petermuessig:


petermuessig:
@uxkjaer : FYI - I updated the playwright dependencies to the latest. For now, I won't make a release, until more changes are aggregated.

Issue #665, opened 2022-12-30
[ui5-tooling-transpile-middleware] not working in combination with [ui5-middleware-stringreplacer]


dominikfeininger:
**Describe the bug**
In case a `*.ts` file contains a string to replace.

When using `ui5-tooling-transpile-middleware` with `ui5-middleware-stringreplacer` the stringreplacer module seems to look on the filesystem for the `*.js` file which cannot be there because it's live-transpiled from `*.ts`. 

**To Reproduce**
1. use both middlewares in ui5-tooling
2. setup a replace in a live-transpiled `*.ts` file eg. `"${project.path}"`

**Expected behavior**
The string `"${project.path}"` was replaced

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser chrome

**Additional context**

After debugging the issue seems to be the resource detection:
https://github.com/ui5-community/ui5-ecosystem-showcase/blob/f2196bc2cba2ba8076ee18506d1c5c3e4d609b92/packages/ui5-middleware-stringreplacer/lib/stringreplacer.js#L127 

Which retunrs `null` -> `Resource not found`.

petermuessig:
Hi @dominikfeininger ,

yes, this indeed a conceptual problem of the middlewares concept. Right now, if a middleware accepts and handles the request, it will process the request just atomically. No other middlewares are considered anymore. Ideally, the middlewares would just create the resources and the the server handle to respond the resource. I think in tooling V3, the server has been changed in that direction or will be changed. Need to verify this the next days. The current middlewares concept is too limited to stack middlewares...

dominikfeininger:
Hi @petermuessig thank you for the fast reply. I guess this is also the reason why `livereload` and `middleware-transpile` not working together?

dominikfeininger:
> Hi @petermuessig thank you for the fast reply. I guess this is also the reason why `livereload` and `middleware-transpile` not working together?

Sry, my bad. PR incoming ;)

petermuessig:
The `livereload` middleware adds some script to the HTML page. The problem with the middlewares is, if one handled the request, no other middleware will be taken into account.

petermuessig:
Rgd. your fixes for the `ui5-middleware-stringreplacer` and the `ui5-tooling-transpile` - both have been published right now: 
 - `ui5-middleware-stringreplacer` 0.5.0
 - `ui5-tooling-transpile` 0.3.4

PR #666, opened 2023-01-02
chore(ts-app): enable live reload for ts sources

dominikfeininger:
Noticed in #665 livereload is not working in this example -> added `ts` files to the watcher.


petermuessig (approved):

petermuessig:
LGTM - will merge it

PR #671, opened 2023-01-04
default exclude images in [ui5-transpile-middleware]

dominikfeininger:
I had trouble when filetypes like images are handled by the transpile middleware. 

So I figured it helps to always exclude them, since there shouldn't be any purpose to transpile images.

### DIFF
--- a/packages/ui5-tooling-transpile/lib/middleware.js
+++ b/packages/ui5-tooling-transpile/lib/middleware.js
@@ -25,7 +25,7 @@ module.exports = async function ({ resources, options /*, middlewareUtil */ }) {
 	config.includes = config.includes || config.includePatterns;
 
 	// never transpile these default exclusion types
-	const defaultExcludes = [".png", "jpeg", ".jpg"];
+	const defaultExcludes = [".png", ".jpeg", ".jpg"];
 	config.excludes = defaultExcludes.concat(config.excludes || config.excludePatterns || []);
 
 	const babelConfig = await createBabelConfig({ configuration: config, isMiddleware: true });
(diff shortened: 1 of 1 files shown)


packages/ui5-tooling-transpile/lib/middleware.js
```diff
@@ -23,7 +23,10 @@ const { createBabelConfig, normalizeLineFeeds } = require("./util");
 module.exports = async function ({ resources, options /*, middlewareUtil */ }) {
 	const config = options?.configuration || {};
 	config.includes = config.includes || config.includePatterns;
-	config.excludes = config.excludes || config.excludePatterns;
+
+	// never transpile these default exclusion types
+	const defaultExcludes = [".png", "jped", ".jpg"];
+	config.excludes = defaultExcludes.concat(config.excludes || config.excludePatterns);
```
petermuessig, line 29: if both `config.excludes` and `config.excludePatterns` are both undefined the concat will add `undefined` - please add another or to add an empty array:
`config.excludes || config.excludePatterns || []`

petermuessig (changes requested):

packages/ui5-tooling-transpile/lib/middleware.js
petermuessig: BTW: there is also a typo: `"jped"` should be  `".jpeg"`, right?

packages/ui5-tooling-transpile/lib/middleware.js
```diff
@@ -23,7 +23,10 @@ const { createBabelConfig, normalizeLineFeeds } = require("./util");
 module.exports = async function ({ resources, options /*, middlewareUtil */ }) {
 	const config = options?.configuration || {};
 	config.includes = config.includes || config.includePatterns;
-	config.excludes = config.excludes || config.excludePatterns;
+
+	// never transpile these default exclusion types
+	const defaultExcludes = [".png", "jpeg", ".jpg"];
```
petermuessig, line 28: Sorry, but the `dot` is still missing for `jpeg`...

petermuessig (changes requested):

petermuessig (approved):

PR #672, opened 2023-01-04
no replace for mime types images, videos in [ui5-middleware-stringreplace]

dominikfeininger:
I had trouble with images when using the stringreplacer middleware.

It seems to change some encoding since the images were loaded with http status 200 but not displayed correctly.

An easy solution seems to exclude `content-type: image/*` and `content-type: video/*`.

### DIFF
--- a/packages/ui5-middleware-stringreplacer/lib/stringreplacer.js
+++ b/packages/ui5-middleware-stringreplacer/lib/stringreplacer.js
@@ -131,6 +131,14 @@ module.exports = function createMiddleware({ resources, options, middlewareUtil 
 			return;
 		}
 
+		// determine charset and content-type
+		const { contentType, charset } = middlewareUtil.getMimeInfo(resource.getPath());
+		// never replace strings in these mime types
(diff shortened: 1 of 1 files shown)


packages/ui5-middleware-stringreplacer/lib/stringreplacer.js
```diff
@@ -131,6 +131,14 @@ module.exports = function createMiddleware({ resources, options, middlewareUtil
 			return;
 		}
 
+		// determine charset and content-type
+		const { contentType, charset } = middlewareUtil.getMimeInfo(resource.getPath());
+		// never replace strings in these mime types
+		if (contentType.includes("image") || contentType.includes("video")) {
```
petermuessig, line 137: a regex would be better but anyhow - change is good! 👍

petermuessig (approved):

dominikfeininger:
Same issue in `ui5-task-stringreplacer`. Lets see if I can provide a fix as well.

PR #683, opened 2023-02-14
fix(ui5-middleware-cfdestionation): support h2 devserver mode

### LINKED ISSUE #641: [ui5-middleware-cfdestination] does not work when serving with --h2
**Describe the bug**
```
Error: write EPROTO 4376790400:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:332:
```

**To Reproduce**
1. Use ui5-middleware-cfdestination
2. Run with `ui5 serve --h2`
3. See error

petermuessig:
Fixes #641


petermuessig:
@piejanssens  - `request` is deprecated and the piping isn't properly working here. This has to do with different transfer-encoding and protocols, so preferring to proxy the content directly...

vobu (approved):
quick glance only, but looking good!

PR #693, opened 2023-03-12
feat(ui5-tooling-transpile): generate d.ts and filter ts files

petermuessig:


petermuessig:
@texttechne @akudev - here my suggestion to include the `d.ts` files and to exclude the `ts` files from the build result. The `ts` files will now always be excluded from the build result but the generation of the `d.ts` files can also be disabled with the setting `skipGenerateDTS`. If ok, I will merge and release tomorrow...

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -31,54 +32,150 @@ module.exports = async function ({ workspace, dependencies, /* taskUtil,*/ optio
 	}
 
 	// TODO: should we accept the full glob pattern as param or just the file pattern?
-	const allResources = await workspace.byGlob(`/**/*${filePatternConfig}`);
-	if (config.transpileDependencies) {
-		// TODO: does transpileDependencies make sense for JavaScript files?
-		const depsResources = await dependencies.byGlob(`/**/*${filePatternConfig}`);
-		allResources.push(...depsResources);
-	}
+	let allResources = await workspace.byGlob(`/**/*${filePatternConfig}`);
```
texttechne, line 35: can be `const`

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -31,54 +32,150 @@ module.exports = async function ({ workspace, dependencies, /* taskUtil,*/ optio
...
+				//traceResolution: true,
+			};
+
+			// Create a Program with an in-memory emit
+			const host = ts.createCompilerHost(options);
+			(host.getCurrentDirectory = function () {
+				return "";
+			}),
+				(host.fileExists = function (file) {
+					if (sourcesMap[file]) {
```
texttechne, line 116: can be shortened: 
```
return sourcesMap[file] ? true : fs.existsSync(file)
```

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -31,54 +32,150 @@ module.exports = async function ({ workspace, dependencies, /* taskUtil,*/ optio
...
+						return true;
+					} else {
+						return fs.existsSync(file);
+					}
+				});
+			host.readFile = function (file) {
+				if (sourcesMap[file]) {
```
texttechne, line 123: can be shortened to one-liner

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -31,54 +32,150 @@ module.exports = async function ({ workspace, dependencies, /* taskUtil,*/ optio
...
+					path: `${fileName}`,
+					string: content
+				});
+				workspace.write(dtsFile);
+			};
+			host.resolveModuleNames = function (moduleNames, containingFile) {
+				const resolvedModules = [];
+				for (const moduleName of moduleNames) {
+					// try to use standard resolution
+					let result = ts.resolveModuleName(moduleName, containingFile, options, {
```
texttechne, line 141: can be `const`

texttechne (approved):

texttechne:
I'm a bit surprised that the source map integration is hard-coded instead of being configurable.
On the one hand, not all people want to publish their source code in this way; usually some will insist on obfuscation.
On the other hand, there are multiple options to generate different kinds of source maps with different impact on build time and debugging quality.

petermuessig:
The source maps integration needs to be hard-coded in this case since we need to create additional files in the workspace - so there needs to be some glue code somewhere. But the decision whether source maps will be created is based on the configuration in the `.babelrc` file. If the source maps are activated there, the transpile function returns an object with the source map attached and this needs to be written into the workspace.

packages/ui5-tooling-transpile/lib/task.js
petermuessig: 👍

packages/ui5-tooling-transpile/lib/task.js
petermuessig: Damn, the auto completion of VSCode completely changed the code and put it into brackets... FYI: the reason why I have chosen those lengthy code was to be able to put break points inside and add additional logging - therefore I prefer to stick to the lengthy code. And the reason why I used classic functions is to be able to use the `arguments` when logging... 😉

packages/ui5-tooling-transpile/lib/task.js
petermuessig: copy paste from here: https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API - not reworked all places properly... 😁

texttechne:
Ahhh, configurable via `.babelrc`. Overlooked, but understood now 😄.

The improvements I've suggested are only minor in nature. I've tested this against ui5-mobx and it works as it should. Hence I already approved this PR (not that it makes any difference 😉).

petermuessig:
Ok, I'll rename

PR #696, opened 2023-03-19
fix: improvements of projects for tooling 3.0

petermuessig:
build: always use chromedriver matching the environment

chore: added simple app using tooling 3.0

fix: fixed the tsconfig.json of TS projects

fix: updated UI5 version

fix(ui5-task-minify-xml): add tooling 3.0 dependency determination

fix(ui5-tooling-modules): only rely on workspace dependencies (to find entry points)

fix(ui5-tooling-modules): improved logging

fix(ui5-task-compileless): fixed dependency determination (theme build requires all deps)


petermuessig:
@flovogt : 
* found out that the less task requires all dependencies, similar like the library build (btw: this is also the reason why the ui5-app runs the build on all dependencies)
* one task was missing: the minify xml task 
* added a very simple app to validate the tooling 3.0 with the two most important middlewares/tasks: ui5-tooling-modules and ui5-tooling-transpile - maybe the proxy would be another interesting one... 
* some more small bugfixes and cleanups

packages/ui5-app-simple/webapp/index.html
```diff
@@ -0,0 +1,25 @@
...
+				"ui5.ecosystem.demo.simpleapp": "./"
+			}'
+			data-sap-ui-async="true"
+			data-sap-ui-frameOptions="trusted"
+			data-sap-ui-language="en"
+		></script>
+	</head>
+	<body class="sapUiBody">
+		<div data-sap-ui-component data-url="./" data-name="ui5.ecosystem.demo.simpleapp" data-id="container" data-settings='{"id" : "Sample"}'></div>
+		<!-- div data-sap-ui-component data-url="./" data-name="ui5.ecosystem.demo.simpleapp" data-id="container" data-settings='{"id" : "Sample"}'></div -->
```
flovogt, line 23: AFAIS This commented code can be removed.

packages/ui5-tooling-modules/lib/middleware.js
```diff
@@ -78,3 +77,19 @@ module.exports = function ({ resources, options, middlewareUtil }) {
 		next();
 	};
 };
+
+//const {parse, print} = require("recast");
```
flovogt, line 81: AFAIS this code can be removed.

flovogt (changes requested):
Hi @petermuessig,
thanks a lot addressing all these points.

> found out that the less task requires all dependencies, similar like the library build (btw: this is also the reason why the ui5-app runs the build on all dependencies)

Ack, Thx for fixing

__________

> one task was missing: the minify xml task

Ack. Thx for fixing

__________

> added a very simple app to validate the tooling 3.0 with the two most important middlewares/tasks: ui5-tooling-modules and ui5-tooling-transpile - maybe the proxy would be another interesting one...

Ack.

__________


> some more small bugfixes and cleanups

Ack.

I have only minor comments

flovogt (approved):
LGTM 🎊

PR #698, opened 2023-03-26
fix(ui5-tooling-modules): allow to define provided dependencies

petermuessig:
Provided dependencies will not be included by the ui5-tooling-modules-task and must be provided from the application runtime environment. Although the tasks ignores provided dependencies, the middleware still serves them.


petermuessig:
@texttechne - here your last requirement -> the option to define `providedDependencies`. With this option you can provide an array of NPM packages which will be handled as provided by the application runtime environment and not be packaged or scanned further for transitive dependencies to be included. Please take a look into this change...

packages/ui5-tooling-modules/README.md
```diff
@@ -46,6 +46,9 @@ npm install ui5-tooling-modules --save-dev
 - *removeScopePreceder*: `boolean`
   Removes the scope preceder `@` from the namespace/path of the 3rd party module when adding it to the namespace with the *addToNamespace* option.
```
akudev, line 47: What is a "preceder"? A "prefix", maybe? :-)

petermuessig: Looking into the [documentation of NPM](https://docs.npmjs.com/cli/v9/using-npm/scope) it is stated like that:

*When used in package names, scopes are preceded by an @ symbol and followed by a slash, e.g.*

*preceded* => **preceder**

But somehow my research wasn't careful enough and it noun **preceder** doesn't exist in English - so'll adopt back to prefix...

akudev (approved):

PR #708, opened 2023-04-02
fix(ui5-tooling-transpile): use proper fs path for resource tranform

petermuessig:
To benefit from out of the box support of other Babel plugins the `filename` needs to point to the proper `fsPath` of the resource. In Tooling 3.0 this works out-of-the-box and for 2.0 we are using a workaround (which will be removed with the full upgrade to UI5 Tooling 3.0).


petermuessig:
Cc: @matz3 @flovogt @RandomByte 

This is needed to benefit from an out-of-the-box support of `babel-plugin-istanbul` for the instrumentation and the reporting back to the proper TypeScript or in general transpiled files. As this work quite OK for UI5 Tooling 3.0 (except of the determination whether the resource is from a root project - which can be maybe considered to be added?) it works quite good. For 2.0 I can live with the workaround as this is nevertheless a dead end...

PR #709, opened 2023-04-03
Livereload for less

dominikfeininger:
To enable a seamless livereload for less files without page reload.

Wasn't able to find another solution than to dig deeper into the used `node_modules` which causes modifications in `livereload-js` which is used as an indirect subdepedency via `node-livereload.

@petermuessig Suggestions welcome.


dominikfeininger:
POC and working example.

Not recommended to merge since depended modules need change as well. 

Evaluation of other livereload libs required.

@vobu I would be happy to fulfill the `Derived Beer-ware License` if you can provide this feature ;)

petermuessig:
Still needed?

PR #734, opened 2023-04-21
fix(ui5-tooling-transpile): d.ts files to be modules + generate index.d.ts

petermuessig:
This fix ensures the correct d.ts generation. The transpile procees now creates an index.d.ts and all TS UI5 modules will declare its module with the fully qualified name:

```ts
declare module "${moduleName}" {
  ${source}
}
```

In addition, the generated control interfaces, the `.gen.d.ts` files need to also be converted to declare the module with the fully qualified name and put the imports into the module declaration.

---

The `ui5-tooling-transpile` tooling extension to properly generate the type definitions including the `index.d.ts` which ensures that e.g. TypeScript UI5 applications consuming UI5 libraries will get code completion by just adding the `typeRoots` configuration:

Snippet taken from: https://github.com/ui5-community/ui5-ecosystem-showcase/blob/223b9f54745c397f6bf7aec430d45d5a9a33c354/showcases/ui5-tsapp/tsconfig.json:

```json
{
  "compilerOptions": {
    [...]
    "paths": {
      [...]
      "ui5/ecosystem/demo/tslib/*": ["./node_modules/ui5-tslib/src/ui5/ecosystem/demo/tslib/*"]
    },
    "typeRoots": [
      "./node_modules/@types",
      "./node_modules/ui5-tslib"
    ]
  },
  "include": ["./webapp/**/*.ts"]
}
```

The `paths` mapping is just included in case of the library hasn't been built yet in the monorepo structure. But this mapping can also be omitted. Important is the addition of the `typeRoots` configuration - keep in mind, as soon as you add your type root for e.g. `node_modules/ui5-tslib`, you need to also add the type roots to `node_modules/@types` so that the types from definitely typed can be found.


showcases/ui5-tsapp/tsconfig.json
```diff
@@ -8,14 +8,14 @@
...
-    "baseUrl": "./",
     "sourceMap": true,
     "paths": {
       "ui5/ecosystem/demo/tsapp/*": ["./webapp/*"],
-      "ui5/ecosystem/demo/tslib/*": ["./node_modules/ui5-tslib/src/ui5/ecosystem/demo/tslib/*"],
       "unit/*": ["./webapp/test/unit/*"],
-      "integration/*": ["./webapp/test/integration/*"]
-    }
+      "integration/*": ["./webapp/test/integration/*"],
+      "ui5/ecosystem/demo/tslib/*": ["./node_modules/ui5-tslib/src/ui5/ecosystem/demo/tslib/*"]
```
akudev, line 16: What' the difference here? Why do we still need a path into node_modules?

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -119,6 +119,24 @@ module.exports = async function ({ workspace /*, dependencies*/, taskUtil, optio
 					//traceResolution: true,
 				};
 
+				// update the sources map to declare the modules with the full module name
+				for await (const resourcePath of Object.keys(sourcesMap)) {
+					const source = sourcesMap[resourcePath];
+					let moduleName = /^\/resources\/(.*)\.ts$/.exec(resourcePath)?.[1];
+					if (moduleName?.endsWith(".gen.d")) {
+						moduleName = /^(.*)\.gen\.d$/.exec(moduleName)?.[1] || moduleName;
```
akudev, line 127: Ummm... isn't the OR part in the end of this line (and the question mark) unneeded? You already checked that the module name ends with ".gen.d".

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -119,6 +119,24 @@ module.exports = async function ({ workspace /*, dependencies*/, taskUtil, optio
 					//traceResolution: true,
 				};
 
+				// update the sources map to declare the modules with the full module name
+				for await (const resourcePath of Object.keys(sourcesMap)) {
+					const source = sourcesMap[resourcePath];
+					let moduleName = /^\/resources\/(.*)\.ts$/.exec(resourcePath)?.[1];
+					if (moduleName?.endsWith(".gen.d")) {
```
akudev, line 126: Hmmm.. so the heuristics is: any module named "*.gen.d" is assumed to be one generated by the ts-interface-generator and hence already containing a module declaration, which has to be modified, though? Should maybe be written down somewhere.

akudev (commented):
Looks good to me overall

showcases/ui5-tsapp/tsconfig.json
akudev: Or is this related to the following statement from the [issue report](https://github.com/SAP/ui5-typescript/issues/382#issuecomment-1517501426)?? 

> "The paths mapping is just included in case of the library hasn't been built yet in the monorepo structure. But this mapping can also be omitted. "

petermuessig: The paths mappings are still necessary as long as the library has not been built. Then this is used to lookup the type information from the `ts` files. Once built, the type definitions are retrieved from the `dist` folder, then the `index.d.ts` and the nested `d.ts` files are used and the `typeRoots` configuration wins.

packages/ui5-tooling-transpile/lib/task.js
```diff
@@ -171,6 +188,23 @@ module.exports = async function ({ workspace /*, dependencies*/, taskUtil, optio
...
 				}
+
+				// create the index.d.ts in the root output folder
+				config.debug && log.info(`Generating index.d.ts`);
+				const indexDtsContent = Object.keys(sourcesMap)
+					.filter((dtsFile) => dtsFile.startsWith("/resources/"))
+					.map(
+						(dtsFile) =>
+							`/// <reference path=".${
+								/\.d\.ts$/.test(dtsFile) ? dtsFile : dtsFile.replace(/\.ts$/, ".d.ts")
```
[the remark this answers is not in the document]
petermuessig: Correct - and for controls there are even 2: the d.ts and the gen.d.ts

packages/ui5-tooling-transpile/lib/task.js
petermuessig: Relict from previous logic - will adapt

packages/ui5-tooling-transpile/lib/task.js
petermuessig: Adding JSDoc

PR #749, opened 2023-05-30
fix: update dependencies

petermuessig:


tobiasqueck:
@petermuessig how do you want me to review this one?
I can see UI5 updates to 1.114 and cds to 6.8.2, these, I understand and they needed to be done. For the others, I would need an explanation.

petermuessig:
@tobiasqueck - it is more the regular update of the dependencies to ensure using the latest versions. It is more for me on an informational level, especially because we release a new version of the `ui5-tooling-transpile` and when you take a look into the `package.json` you can see a new version of the Babel plugin to transform the UI5 modules: `7.2.0`. The changelog can be found here: https://github.com/ui5-community/babel-plugin-transform-modules-ui5/blob/main/CHANGELOG.md

PR #750, opened 2023-06-01
fix(cds-plugin-ui5): disable gzip encoding when livereload is used by CAP

petermuessig:


tobiasqueck:
@petermuessig I have tested it locally and it works without problems!

PR #765, opened 2023-07-14
feat(ui5-middleware-cfdestination): allow dest. to be maintained in .env

stockbal:
Provides the option to store destinations as JSON string inside in `.env` file variable. 

Extract from `ui5.yaml` that is using the new option:

```yaml
      afterMiddleware: compression
      configuration:
        authenticationMethod: "route"
        allowServices: true
        debug: true
        xsappJson: "local-xs-app.json"
        destinations: "$env:xsapp_dest"
```

P.S.
As I am new to `ava` which is used for testing I only wrote a positive test. I didn't manage it to also implement a negative test 😊.

Best regards,
Ludwig

### DIFF
--- a/packages/ui5-middleware-cfdestination/lib/cfdestination.js
+++ b/packages/ui5-middleware-cfdestination/lib/cfdestination.js
@@ -103,11 +103,15 @@ module.exports = async ({ resources, options, middlewareUtil }) => {
 
 	// check if destinations exist in .env file as JSON string
 	if (typeof effectiveOptions.destinations === "string" && effectiveOptions.destinations.startsWith("$env:")) {
-		const destinationsEnvKey = effectiveOptions.destinations.substring(5)
-		try {
-			effectiveOptions.destinations = JSON.parse(process.env[destinationsEnvKey])
-		} catch (error) {
-			throw new Error(`No valid destinations JSON in .env file at '${destinationsEnvKey}': ${error}`)
+		const destinationsEnvKey = effectiveOptions.destinations.substring(5).trim()
+		if (destinationsEnvKey && destinationsEnvKey in process.env) {
+			try {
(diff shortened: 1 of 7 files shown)


packages/ui5-middleware-cfdestination/lib/cfdestination.js
```diff
@@ -99,6 +101,19 @@ module.exports = async ({ resources, options, middlewareUtil }) => {
 	// the default auth mechanism is set to none but the user can pass an auth method using the options
 	xsappConfig.authenticationMethod = effectiveOptions.authenticationMethod
 
+	// check if destinations exist in .env file as JSON string
+	if (typeof effectiveOptions.destinations === "string" && effectiveOptions.destinations.startsWith("$env:")) {
```
petermuessig, line 105: I personally would prefer a regex for those kind of checks but this is also good to go!

packages/ui5-middleware-cfdestination/test/dest-in-env/ui5.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: themelib_sap_horizon
+server:
+  customMiddleware:
+  - name: ui5-middleware-cfdestination
+    afterMiddleware: compression
+    configuration:
+      debug: true
+      port: 1091
+      xsappJson: "xs-app.json"
+      destinations: "$env:xsapp_dest"
```
petermuessig, line 21: I always thought of a general concept to load properties for the `ui5.yaml` from an env file to standardize this. Similar discussions we had for proxies and their username/password and we do similar things for the stringreplace middleware/task.

Cc @RandomByte @matz3 - something to think about for the evolution of the UI5 Tooling

petermuessig (approved):
LGTM overall, thanks for your contribution!

petermuessig:
Version `0.11.0` is about to be released!

PR #779, opened 2023-07-31
fix: more cleanup of obsolete deps

petermuessig:


flovogt:
LGTM but there might be lock file issues

petermuessig:
The only packages still using dependencies to the `@ui5/fs` are the tooling extensions for `less`. They also make use of the `fsInterface`. Not sure yet, what to do with them...

Cc: @RandomByte @matz3

flovogt (commented):
LGTM

PR #787, opened 2023-08-07
feat(ui5-tooling-transpile): integrate ts-interface-generator

petermuessig:
If a TypeScript project adds the dependency to the ts-interface-generator the ui5-tooling-transpile-middleware tries to start the watch mode so that changes to control files will automatically update the interfaces and that code completion can work properly. For now this only works for the local project, not for dependencies.


petermuessig:
⚠️ I think, I will first enhance the `@ui5/ts-interface-generator` to have a proper API and error handling when running embedded in another module. And I need to enhance the detection of the local dependency to `@ui5/ts-interface-generator` to be a `dependency` or `devDependency` and not transitive...

PR #788, opened 2023-08-08
feat: Apply `instrumenterOptions` for Istanbul from `coverageReporter` in `karma.config`

MuwuM:
This pull request addresses an issue with code coverage reporting when running Karma tests for UI5 in an iframe. Previously, coverage was not being correctly reported in such scenarios.

`karma-ui5` [provides a solution to this problem](https://github.com/SAP/karma-ui5#configureiframecoverage) by enabling code coverage for iframes. This is done through the use of the `configureIframeCoverage` helper function, which should be called from the Karma configuration function after the coverage plugin has been configured.

With the changes introduced in this PR, the `karma-ui5-transpile` will now correctly apply these configurations, ensuring accurate coverage reporting when running Karma tests for UI5 applications within an iframe.

### DIFF
--- a/packages/karma-ui5-transpile/lib/index.js
+++ b/packages/karma-ui5-transpile/lib/index.js
@@ -9,7 +9,6 @@ const yaml = require("js-yaml");
  *
  * Kudos goes to: https://github.com/babel/karma-babel-preprocessor
  * which inspired the creation of this Karma preprocessor.
- *
  * @param {object} config Config object of UI5.
  * @param {object} logger Karma's logger.
  * @returns {Function} The preprocess function.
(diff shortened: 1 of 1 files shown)


packages/karma-ui5-transpile/lib/index.js
```diff
@@ -75,21 +74,23 @@ function createPreprocessor(config, logger) {
 				return plugin.file.request === "istanbul";
 			})
 		) {
+			const istanbulConfig = {
+				include: ["**/*"],
+				exclude: []
+			};
+			// apply the `instrumenterOptions` for istanbul from `coverageReporter` in the `karma.config`
+			const instrumenterOptionsIstanbul = config.coverageReporter?.instrumenterOptions?.istanbul;
+			if (instrumenterOptionsIstanbul && typeof instrumenterOptionsIstanbul === "object") {
```
petermuessig, line 83: A simple `typeof` check is sufficient in this case as `typeof undefined === "object"` => `false`

packages/karma-ui5-transpile/lib/index.js
```diff
@@ -75,21 +74,23 @@ function createPreprocessor(config, logger) {
...
+				Object.assign(istanbulConfig, config.coverageReporter.instrumenterOptions.istanbul);
```
petermuessig, line 84: Here you could use `instrumenterOptionsIstanbul` instead of `config.coverageReporter.instrumenterOptions.istanbul`

petermuessig (changes requested):
Two minor things

packages/karma-ui5-transpile/lib/index.js
MuwuM: Yes, I somehow had the misconception in my head, that optional chaining would yield `null` instead of `undefined`. Checked documentation and it looks like I was wrong. (maybe I got that from some strange polyfills I used for it in the past)

packages/karma-ui5-transpile/lib/index.js
MuwuM: That was actually the reason for creating `instrumenterOptionsIstanbul` as a variable ... seems like I have been too tired yesterday

petermuessig (approved):

PR #792, opened 2023-08-12
fix(ui5-tooling-modules): allow to skip convert of dynamic imports

petermuessig:
If a project needs to consume an NPM package which makes use of dynamic imports to load arbitrary scripts, one can decide to keep the `import(...)` call and omit the transpiling into AMDs' `require(...)` call.

This is e.g. necessary for `@luigi-project/container` to support loading custom elements as ES modules from another location. To do so, you need to do the following:

* Add a dependency to `@luigi-project/container`
* Extend the `ui5-tooling-modules` configuration to tell it to keep dynamic imports for this module
  * Example to define reuse configuration centrally in `ui5.yaml`: https://github.com/ui5-community/ui5-ecosystem-showcase/blob/3675ed95b8f6408fd9fe9e36a4fc5aba9ec10518/showcases/ui5-tsapp/ui5.yaml#L10-L13 and reuse it in the task and middleware)
  * Reuse in task: https://github.com/ui5-community/ui5-ecosystem-showcase/blob/3675ed95b8f6408fd9fe9e36a4fc5aba9ec10518/showcases/ui5-tsapp/ui5.yaml#L37-L40
  * Reuse in middleware: https://github.com/ui5-community/ui5-ecosystem-showcase/blob/3675ed95b8f6408fd9fe9e36a4fc5aba9ec10518/showcases/ui5-tsapp/ui5.yaml#L52-L55
* Now the generated file keeps the dynamic import: `import(...)`

### DIFF
--- a/packages/ui5-tooling-modules/README.md
+++ b/packages/ui5-tooling-modules/README.md
@@ -37,9 +37,21 @@ npm install ui5-tooling-modules --save-dev
 
 ## Configuration options (in `$yourapp/ui5.yaml`)
 
+The following configuration options are relevant for the `task` and the `middleware`:
+
 - *verbose*: `boolean`  
   Enables verbose logging (defaults to `false`)
   &nbsp;
+
+- *skipCache*: `boolean` *experimental feature*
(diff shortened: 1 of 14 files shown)


petermuessig:
@hardl - this is my recommendation! WDYT?

PR #800, opened 2023-08-21
feat(cds-plugin-ui5): allow to attach custom app pages to cds server

### LINKED ISSUE #785: [cds-plugin-ui5] How to show virtual html links?
**Is your feature request related to a problem? Please describe.**
As part of the Open UX tools, a new middleware was introduced that exposes an FLP sandbox on-the-fly for previewing an application in an FLP. With this middleware, the project won't have a static html file anymore as it was in the past. Using the `preview-middleware` works with the `cds-plugin-ui5` i.e. the html file is returned as expected.
However, the file is (obviously) not listed when I just open `http://localhost:4004` because the `cds-plugin-ui5` does not know that the `preview-middleware` is exposing the html file on-the-fly.

**Describe the solution you'd like**
I would like to see the on-the-fly generate html in the root html page generated/enhanced by the `cds-plugin-ui5`
How? I don't think the `cds-plugin-ui5` can "magically" know if there is an html file hosted on-the-fly, so some kind of hook or well defined API is needed that the `preview-middleware` could implement

**Additional context**
My test project is at https://github.com/tobiasqueck/ui5con2023/tree/preview-middleware, it is a fork of the demo that @MariusFreitag gave at UI5Con2023.

@petermuessig here is the problem that I mentioned in our call

petermuessig:
If custom middlewares assign a static `getAppPages` function to their middleware function, the app pages returned by that function will be added to the welcome page of the `@sap/cds` server. This enables middlewares providing virtual pages to list them in the welcome page.

Fixes #785

### DIFF
--- a/packages/cds-plugin-ui5/README.md
+++ b/packages/cds-plugin-ui5/README.md
@@ -21,6 +21,34 @@ npm add cds-plugin-ui5 -D
 
 That's it!
 
+## Info for UI5 Tooling Extension Developers
+
+Custom middlewares may generate virtual app pages which should also be listed as web applications in the welcome page of the `@sap/cds` server. This is possible by assigning a static `getAppPages` function to the middleware function. The following snippet show-cases how this can be done:
+
+```js
+// the middleware factory function
+module.exports = async ({ log, resources, options }) => {
+  // create the middleware function
+  const mwFn = (req, res, next) => {
+    [...]
+  };
+
+  /**
+   * Returns an array of app pages to be added to the welcome
+   * page of the <code>@sap/cds</code> server.
+   * @returns {undefined|string[]} array of app pages
+   */
+  mwFn.getAppPages = function() {
+    return ["/test.html"];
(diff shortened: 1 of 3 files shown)


petermuessig:
@nicoschoenteich - this also requires changes in the `dev-approuter`. The app pages returned by the `applyUI5Middleware` util are now an array of strings instead an array of Resources. As soon as the new feature version is being released and you upgrade you need to consider this.

nicoschoenteich:
Hi @petermuessig, thanks for the info. I just checked and noticed that we don't even use the return value of `applyUI5Middleware` in the `dev-approuter` - so no changes required.

tobiasqueck:
@petermuessig just to be sure that I understand your proposal correctly, for our `@sap-ux/preview-middleware` I would just have to do https://github.com/SAP/open-ux-tools/compare/poc/cds-plugin-ui5?expand=1? If yes, then, I like the solution.

PR #805, opened 2023-08-25
fix(ui5-tooling-transpile): use latest UI5 Tooling V3 APIs only

petermuessig:


petermuessig:
@RandomByte @matz3 - just FYI - this finishes the migration to UI5 Tooling V3 for the ui5-tooling-transpile... ;-)

PR #819, opened 2023-08-31
feat(ui5-middleware-websocket): enable websocket support for UI5 server

### LINKED ISSUE #557: [ui5-middleware-cfdestination] need websockets support
We were trying to use a local websocket mock, which works fine, when it's directly used by the frontend. But we were unable to use this with ui5-middleware-cfdestination.

I did try to enable ws through the approuter xsappConfig like this:
```
    xsappConfig.websockets = {
        "enabled": true
    };
```
But I did have to add the additional ports to WS_ALLOWED_ORIGINS to get it working:
```
    process.env.WS_ALLOWED_ORIGINS = JSON.stringify([
        {
            "protocol": "http",
            "host": "localhost",
            "port": 8082
        }
    ]);
```
AND add express-ws to the internally used express server:
```
    expressWS = require('express-ws')(req.app);
```

This will forward the initial websocket UPGRADE request from the client to the mock server (on port 8082). Unfortunately it still does not work, as the line `req.pipe(request(url)).pipe(res)` only works for requests (not for continuous websockets). 

One possible solution could be to check the req method and use something else than a request to pipe (proxy) the request object to our mock api. This would still use approuter, but we would have to traverse destinations and add them to WS_ALLOWED_ORIGINS programmatically. And maybe other hacks.

An alternative could be to use a more uncomplicated router project like [http-proxy-middleware](https://www.npmjs.com/package/http-proxy-middleware), I tried a quick experiment and it seems to work:
```
    // can be filled from destinations
    const proxyTable = {
        '/CRM_BACKEND_WS': 'http://localhost:8082', // path only
        '/CRM_BACKEND': 'http://localhost:8081' // path only
    };
  
    const proxyOptions = {
        target: 'http://localhost:8080',
        ws: true,
        router: proxyTable,
        logLevel: 'debug',
        proxyTimeout: 500
    };
    
    const myProxy = createProxyMiddleware(proxyOptions);

    return (req, res, next) => {    
        const app = req.app
        app.use(myProxy); // add the proxy to express

        next()
    }
```

petermuessig:
This commit introduces a new middleware package `ui5-middleware-websocket`
which provides helper functions to create WebSocket middlewares inside the
UI5 server:

```js
const websocket = require("ui5-middleware-websocket/lib/websocket");

/**
 * WebSocket middleware to act as an echo service.
 *
 * @param {object} parameters Parameters
 * @param {@ui5/logger/Logger} parameters.log Logger instance
 * @returns {Function} Middleware function to use
 */
module.exports = ({ log }) => {
  return websocket(function echo(ws, req /*, next */) {
    ws.on("message", function (message) {
      log.info(`message: ${message}`);
      ws.send(`echo ${message}`);
    });
    log.info(`Connection established with ${req.url}`);
  });
};
```

The function `websocket` provided by the module:
`const websocket = require("ui5-middleware-websocket/lib/websocket");`
can be used in middlewares to install a WebSocket callback function.

In addition, the middleware package provides an WebSocket echo service
which can be installed for testing in your UI5 server:

```yaml
server:
  customMiddleware:
  - name: ui5-middleware-websocket-echo
    afterMiddleware: compression
    mountPath: /wsecho
```

This allows to test basic WebSocket functions, e.g. by using the
`sap.ui.core.ws.WebSocket` helper (see `Main.controller.js`).

All proxies `ui5-middleware-simpleproxy` and `ui5-middleware-cfdestination`
have been enabled to support proxying WebSockets. The functionality has
to be enabled by a configuration option `enableWebSocket`:

```yaml
    - name: ui5-middleware-simpleproxy
      mountPath: /spwsecho
      afterMiddleware: compression
      configuration:
        debug: true
        baseUri: "ws://localhost:1081/wsecho/"
        enableWebSocket: true
      afterMiddleware: compression
      configuration:
        port: 1091
        xsappJson: "xs-app.json"
        destinations:
          - name: "cfwsecho"
            url: "ws://localhost:1081/wsecho/"
        enableWebSocket: true
```

Bonus: tests have been added which showcases the usage of `superwstest` to
test the websocket functionality in both middlewares.

Fixes https://github.com/ui5-community/ui5-ecosystem-showcase/issues/557

### DIFF
--- a/packages/ui5-middleware-websocket/ui5.yaml
+++ b/packages/ui5-middleware-websocket/ui5.yaml
@@ -1,0 +1,8 @@
+# https://sap.github.io/ui5-tooling/pages/extensibility/CustomServerMiddleware/
+specVersion: "3.0"
+metadata:
+  name: ui5-middleware-websocket-echo
+kind: extension
+type: server-middleware
+middleware:
+  path: lib/echo.js
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ packages
 ├── ui5-middleware-servestatic      // middleware extension: serve static resources
 ├── ui5-middleware-simpleproxy      // middleware extension: simple express proxy
 ├── ui5-middleware-webjars          // middleware extension: deliver content from JAR files
+├── ui5-middleware-websocket        // middleware extension: enable web sockets for the UI5 Tooling
 ├── ui5-task-cachebuster            // task extension: enables cachebusting for standalone applications
 ├── ui5-task-compileless            // task extension: compile less files in the app folder
 ├── ui5-task-flatten-library        // task extension: prepares build result for deployment to SAP NetWeaver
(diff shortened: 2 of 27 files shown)


packages/ui5-middleware-websocket/ui5.yaml
```diff
@@ -0,0 +1,16 @@
...
+middleware:
+  path: lib/index.js
+---
+specVersion: "3.0"
+metadata:
+  name: ui5-middleware-websocket-test
+kind: extension
+type: server-middleware
+middleware:
+  path: lib/test.js
```
matz3, line 16: This exposes the `ui5-middleware-websocket-test` middleware for consumers of the package, right?
What about moving it to another file?

petermuessig:
Topics to be done:
* README.md => echo middleware!
* Main.controller.js => make example work for cfdestination, simpleproxy and echo middleware

PR #821, opened 2023-09-01
fix(ui5-middleware-cfdestination): make ws configurable

petermuessig:


petermuessig:
@piejanssens : FYI - the `ws` option may have some side effects, therefore I make it configurable until I understand it - and I enabled the option `followRedirects` - just as a try...

PR #823, opened 2023-09-01
fix(cds-plugin-ui5): exclude html fragments from app pages list

BenReim:


petermuessig:
Another good catch! Will try to make a single glob for that!

petermuessig:
`**/!(*.fragment).{html,htm}` does what you want... 😄

Issue #817, opened 2023-08-30
[ui5-middleware-cfdestination] UAA issues


piejanssens:
- in case UAA session has timed out, UI5 tooling (8080) will just report 401 errors, while the real approuter process (5000/5000) would kickoff re-auth process

Related @sap/approuter docu: https://www.npmjs.com/package/@sap/approuter#session-handling

- in case a new UAA session is required (upon re-running ui5 serve, the @sap/approuter will send a redirect to the callback on the approuter port after which you manually have to change the URL back to '8080'

As noted by @petermuessig in https://github.com/ui5-community/ui5-ecosystem-showcase/issues/814#issuecomment-1698107415, this could be resolved after switching to `http-proxy-middleware`

- UI5 serve is not proxying /login/callback to @sap/approuter

petermuessig:
Should be fixed with `ui5-middleware-cfdestination` version `3.1.0` which will be released just right now! Please validate and let me know... THX!

piejanssens:
Hi @petermuessig,

The auth flow is not kicked off any longer. 

> If there is no session in the application router, either because there has been a session timeout or because no session has been created yet, and if the incoming request matches a non-public route, the application router triggers a redirect to the authentication service (UAA or IAS). After a successful login, a redirect back to application router takes place using the login callback endpoint, which triggers the creation of a new session. If the incoming request is an AJAX request (has the request header X-Requested-With: XMLHttpRequest) or if the HTTP verb is not GET and no session exists (there is no session cookie and the request doesn’t have an x-approuter-authorization header), the application router returns the response code 401 - Unauthorized. This enables the client application to handle the 401 response before it navigates to the authentication service. For example, the application can store data entered by the user and prevent data loss. When the handling of the 401 response is completed, the client application should send a request without an xmlhttprequest object to trigger the application router authentication flow.

OData requests performed by UI5 include the header X-Requested-With: XMLHttpRequest. I'm seeing 401 being returned by the approuter, which is correct according to the documentation. The problem I think is that there is no route being accessed on the approuter that does not contain this header (e.g. no request to HTML5 repo service or a static resource protected by XSUAA), which would have gotten a valid session before UI5 is accessing XSUAA procted resources.

This means for `ui5 serve` it's '/' that should be a route handled by @sap/approuter and considerd a non-public route.
As a hack around this I tried adding this route in the xs-app.json:

```
    {
      "source": "^/",
      "localDir": "webapp",
      "target": "/",
[...]

PR #831, opened 2023-09-04
fix(ui5-middleware-cfdestination): keep content-type parameters intact

piejanssens:
Fixes: #817 

P.S. I'm not familiar with the reason that content-type should ever be changed here or defaulted to `application/octet-stream`, so I've added this fix while keeping the original fallback logic.

### DIFF
--- a/packages/ui5-middleware-cfdestination/lib/cfdestination.js
+++ b/packages/ui5-middleware-cfdestination/lib/cfdestination.js
@@ -187,11 +187,11 @@ module.exports = async ({ log, options, middlewareUtil }) => {
 	}
 
 	// helper to determine the mime type either from the req path or if provided
-	// we parse the content-type value and retunr content type and charset
+	// we parse the content-type value and return the content-type and charset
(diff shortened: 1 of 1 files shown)


petermuessig:
The change is producing a wrong content type: 

`'application/xml;charset=utf-8; charset=utf-8'` 

It is more interesting why the $batch hasn't the proper mime type from the original server - because this content type is parsed!

petermuessig:
Ok got it - $batch requests send: 

`multipart/mixed; boundary=batch_id-1693905819494-15`

This needs to be treated properly

petermuessig:
The problem is something else - the `getMimeInfo` doesn't consider the `multipart/mixed` with additional `parameters`, e.g. the `boundary` string which is necessary to split the `$batch` request. Wait a second, I push my change...

petermuessig:
@piejanssens #832 takes over the right `content-type` and is not changing it - only in case of a `content-type` is missing, it will determine it via the `MiddlewareUtil` from the request path

piejanssens:
OK

PR #883, opened 2023-10-14
feat(ui5-tooling-stringreplace): allow combination with other middlewares

### LINKED ISSUE #665: [ui5-tooling-transpile-middleware] not working in combination with [ui5-middleware-stringreplacer]

petermuessig:
The stringreplace middleware must not be the last middleware in the chain and therefore should only intercept the response rather than sending the response. This allows other middlewares such as the transpiling to run afterwards.

The middleware must be executed before any other middleware which finally sends the response (like the transpile or modules middleware). When called, it hooks into the functions of the server response and collects the content to finally rewrite it. The final send will be left to the UI5 tooling development server or other middlewares.

Fixes #665

### DIFF
--- a/packages/ui5-tooling-stringreplace/lib/middleware.js
+++ b/packages/ui5-tooling-stringreplace/lib/middleware.js
@@ -1,16 +1,9 @@
 /* eslint-disable no-undef */
-const { createReplacePlaceholdersDestination, addPlaceholderString, getMimeInfo, readPlaceholderFromEnv, isPathOnContentTypeExcludeList } = require("./util");
+const { createReplacePlaceholdersDestination, addPlaceholderString, readPlaceholderFromEnv, isPathOnContentTypeExcludeList } = require("./util");
 
+const { Readable } = require("stream");
 const minimatch = require("minimatch");
-const etag = require("etag");
-const fresh = require("fresh");
-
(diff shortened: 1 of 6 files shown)


petermuessig:
Cc @dominikfeininger

petermuessig:
Release will follow a bit later today

PR #885, opened 2023-10-16
fix(ui5-tooling-stringreplace): do not ignore transpiled resources

petermuessig:


marianfoo (approved):
Works in my Setup
LGTM!

Issue #901, opened 2023-10-30
Jest test crash when using `cds-plugin-ui5`


js1972:
When using `cds-plugin-ui5` any jest tests in my project fail with a segment fault ass below:

```
*[main][~/dev/consumptionopps]$ npm test     

> consumptionopps@1.0.0 test
> jest tests/test.js

  console.warn
    [cds] - WARNING: 
    
      Returning an 'activate' function is deprecated and won't be
      supported in future releases. → Please return a Promise with 'module.exports' instead.
      

      at Function.logger.warn (node_modules/@sap/cds/lib/log/cds-log.js:98:59)
      at node_modules/@sap/cds/lib/plugins.js:35:26
          at Array.map (<anonymous>)
      at Object.<anonymous>.exports.activate (node_modules/@sap/cds/lib/plugins.js:30:47)
      at cds.get plugins [as plugins] (node_modules/@sap/cds/lib/index.js:10:64)
      at Object.serve (node_modules/@sap/cds/bin/serve.js:159:13)
      at Object.exec (node_modules/@sap/cds/bin/cds-serve.js:14:17)
      at cds.exec (node_modules/@sap/cds/bin/cds-serve.js:50:47)
      at Object.<anonymous> (node_modules/@sap/cds/lib/utils/cds-test.js:57:39)

  console.debug
    [cds-ui5-plugin] [debug] activate {
      impl: '/Users/I540401/dev/consumptionopps/node_modules/cds-plugin-ui5/cds-plugin.js'
    }

      at Object.log (node_modules/cds-plugin-ui5/lib/log.js:20:16)
          at Array.map (<anonymous>)

  console.debug
    [cds-ui5-plugin] [debug] bootstrap

      at Object.log (node_modules/cds-plugin-ui5/lib/log.js:20:16)

  console.info
    [cds-ui5-plugin] [info] Mounting /fre.consumption.opportunities to UI5 app /Users/I540401/dev/consumptionopps/app/opportunities (id=opportunities)

      at Object.log (node_modules/cds-plugin-ui5/lib/log.js:20:16)


 RUNS  tests/test.js
[1]    6167 segmentation fault  npm test
*[main][~/dev/consumptionopps]$ 
```

If I remove the plugin, the test then work properly. Add the plugin again and they fail.

Sample repo: https://github.tools.sap/BTP-CEE/consumptionopps (branch main - Peter you have been added as a contributer).

petermuessig:
[...]

PR #903, opened 2023-10-30
fix: allow for jest runs with plugin

vobu:
a starting point eventually

### DIFF
--- a/packages/cds-plugin-ui5/lib/applyUI5Middleware.js
+++ b/packages/cds-plugin-ui5/lib/applyUI5Middleware.js
@@ -32,8 +32,18 @@ const log = require("./log");
  * @returns {UI5AppInfo} UI5 application information object
  */
 module.exports = async function applyUI5Middleware(router, options) {
-	const { graphFromPackageDependencies } = await import("@ui5/project/graph");
-	const { createReaderCollection } = await import("@ui5/fs/resourceFactory");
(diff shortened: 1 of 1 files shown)


petermuessig (changes requested):
This doesn't work - the modules are ES modules and need to be imported with the dynamic import... Right now this causes an issue...

vobu:
yeah, i figured. this PR was essentially just a reminder to get in done. i‘ll work on it.

petermuessig:
@vobu - the issue is a bit more complicated as the UI5 tooling is built with ESM and therefore can't be required

---

This can't be fixed easily in the `cds-plugin-ui5` as the issue is related to `jest` and `nodejs` - jest forwarded the issue to node:

Jest issue (one of them): https://github.com/jestjs/jest/issues/12286
Node issue: https://github.com/nodejs/node/issues/35889 (fixed with Node.js >= 21)

After upgrading my local nodejs to 21 and using the following command line:
```sh
NODE_OPTIONS=--experimental-vm-modules jest tests/test.js
```

it works... 

This needs to be fixed by using a newer Node.js version plus enabling the ESM support for JEST: https://jestjs.io/docs/ecmascript-modules

Maybe something to add to the documentation... So, I'll keep this PR open for now...

petermuessig:
Closing in favor of #901

PR #951, opened 2024-02-18
fix(cds-plugin-ui5): only start when using "cds serve|watch"

### LINKED ISSUE #950: [cds-plugin-ui5] Plugin also start when mocking a single service
**Describe the bug**
When using the mock function from cds:
https://cap.cloud.sap/docs/guides/using-services#mock-remote-service-as-odata-service-node-js

The cds-plugin-ui5 also loads.

I would have assumed that the mock function should only provide the "backend ODATA service".
I don't know if there is a "flag" or setting to ident that only the mocking is requested.

**To Reproduce**
1. start a "remote" service via "cds mock <servicename>"
2. beside the cds server also the ui5 plugin is startet

**Expected behavior**
When mocking a single service only "backend" function should be loaded

petermuessig:
Fixes #950

### DIFF
--- a/packages/cds-plugin-ui5/cds-plugin.js
+++ b/packages/cds-plugin-ui5/cds-plugin.js
@@ -68,119 +68,125 @@ if (!skip) {
 	cds.on("bootstrap", async function bootstrap(app) {
 		log.debug("bootstrap");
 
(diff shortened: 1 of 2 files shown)


petermuessig:
E2E Test issue is not related

PR #958, opened 2024-02-21
fix(ui5-task-copyright): align env var names

petermuessig:


petermuessig:
E2E issues not related

PR #959, opened 2024-02-23
fix(ui5-tooling-transpile): support special rhino target via env var

petermuessig:


petermuessig:
E2E test issues are not related

PR #969, opened 2024-03-04
chore(cds-plugin-ui5): showcase formatter usage in TS running in CDS

petermuessig:


petermuessig:
Merging the example - to have a proper showcase running

PR #970, opened 2024-03-04
fix(ui5-tooling-transpile): support JS-based babel config files

### LINKED ISSUE #966: [ui5-tooling-transpile] Support babel javascript configuration files
I would like to set a function for the [modulesMap property of the `transform-modules-ui5`](https://github.com/ui5-community/babel-plugin-transform-modules-ui5?tab=readme-ov-file#imports) babel plugin. 
Thus, I have defined a `babel.config.js` file, but this leads the ui5 build to crash when running the `ui5-tooling-transpile` task.

petermuessig:
Fixes #966

### DIFF
--- a/packages/ui5-tooling-transpile/lib/util.js
+++ b/packages/ui5-tooling-transpile/lib/util.js
@@ -63,9 +63,11 @@ async function loadBabelConfigOptions(babelConfigOptions, skipBabelPresetPluginR
 	const partialConfig = await babel.loadPartialConfigAsync(
 		Object.assign(
 			{
+				cwd,
 				configFile: false,
 				babelrc: false,
-				filename: "src/dummy.js" // necessary for ignore/include/exclude
+				filename: "src/dummy.js", // necessary for ignore/include/exclude
+				envName: process.env.BABEL_ENV || process.env.NODE_ENV || "development"
 			},
 			babelConfigOptions
 		)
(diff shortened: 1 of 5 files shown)


BenReim:
Thanks for the fix, Peter! 🎉

packages/ui5-tooling-transpile/lib/util.js
```diff
@@ -112,9 +114,17 @@ async function findBabelConfigOptions(cwd) {
...
-		babelConfigOptions = JSONC.parse(fs.readFileSync(configFile, { encoding: "utf8" }));
-		// let Babel lookup the configuration file with the Babel API
-		//partialConfig = await loadBabelConfigOptions({ configFile, filename: dir, babelrc: true });
+		try {
+			babelConfigOptions = JSONC.parse(fs.readFileSync(configFile, { encoding: "utf8" }));
+		} catch (err) {
+			// no JSON so we let Babel lookup the configuration file with the Babel API
+			const partialConfig = await babel.loadPartialConfigAsync({ cwd });
+			// but we only extract the presets and plugins and ignore the other properties
+			babelConfigOptions = {
```
BenReim, line 123: @petermuessig Is it really intended to ignore other babel config options, such as the `targets` property?
The docu of the task seems to suggest otherwise: 
>The following configuration options will only be taken into account if no inline babel configuration is maintained in the ui5.yaml as babelConfig or no external babel configuration exists in any configuration file as described in [Babels configuration section](https://babeljs.io/docs/configuration)

petermuessig: Well, if I enable more, I need to rework the util.js a bit - will require a bit more time then...

Issue #979, opened 2024-03-26
[ui5-tooling-stringreplace] .ts files not detected for replacing


KevinC0210:
When using the ui5-tooling-stringreplace-task to replace values in files, TypeScript files always get ignored even when adding the .ts ending to the files configuration in the ui5.yaml. 
I looked into it and the issue is the MIME type of the ending .ts. In that case the MIME type results in video/mp2t. As video and image files are excluded from the stringreplace task, the TypeScript files are excluded as well.

When using the ui5-tooling-stringreplace the expected functionality is that it is also usable for replacing strings in TypeScript files.
The easiest fix for this issue would be to make an exception for .ts files when fetching the MIME type so that it does not use the MIME type video/mp2t which is provided by the mime type library.

petermuessig:
I'll release a new version now - thanks for your finding and the PR! 🚀

PR #980, opened 2024-03-26
Adjusted ui5-tooling-stringreplace to work with .ts files

KevinC0210:
I adjusted fetching the MimeInfo fetching as currently the type for all .ts files will result in "video/mp2t" and therefore will be ignored by the stringreplace-task. 
With this change the stringreplace-task can also detect and replace strings in any TypeScript files.

Please also refer to this issue: https://github.com/ui5-community/ui5-ecosystem-showcase/issues/979


petermuessig:
You are right, TypeScript isn't detected properly - but maybe JavaScript is not the proper type and we should use "text/plain" or and custom mime type like "application/x-typescript" - but I'll correct it in a separate change. THX for the FIX 🚀

PR #982, opened 2024-03-27
Correct relative path on offline copy backup

lucawun:


lucawun:
image

petermuessig:
Good catch! Test issues are not related to your change. Seems that the OData services from Microsoft have some issues today. Need to think of mocking them in the future...

petermuessig:
New version `ui5-task-pwa-enabler` => `3.0.7` is about to be released in some mins...

PR #983, opened 2024-03-27
Improve Error-Handle for not in the cache found files

lucawun:
Otherwise, is any unfound files in caching expression:
image


petermuessig:
Nice improvement, this is indeed helpful! Test issues are not related to your change. Seems that the OData services from Microsoft have some issues today. Need to think of mocking them in the future...

petermuessig:
New version `ui5-task-pwa-enabler` => `3.0.7` is about to be released in some mins...

PR #984, opened 2024-03-30
fix: cleanup of dependencies and improved approuter logic

petermuessig:
fix(cds-plugin-ui5): avoid issue when cds is not available

fix(dev-approuter): ensure usage of proper ports for CDS and AppRouter

fix(ui5-middleware-ui5): added dependency to express to allow require


petermuessig:
Test issue related to a bug of the OData services of Microsoft:
https://services.odata.org/V2/Northwind/Northwind.svc/Categories

PR #985, opened 2024-03-31
fix(ui5-tooling-modules): ignore unkown modules for bundling

### LINKED ISSUE #977: ui-tooling-modules 3.3.0 update breaks (typescript) build
**Describe the bug**
After updating the dependency
`"ui5-tooling-modules": "3.2.7"` to `"ui5-tooling-modules": "3.3.0"` the build starts to fail with
```
Error Message:
Resource /resources/chart.js/auto.js has no content
```

**To Reproduce**
customtasks looks like this:
```
  customTasks:
    - name: ui5-tooling-transpile-task
      afterTask: replaceVersion
    - name: ui5-tooling-modules-task
      afterTask: ui5-tooling-transpile-task
```
With `"ui5-tooling-modules": "3.2.7"` it compiles correctly while and copying the dependency `chart.js` to `\dist\resources\chart.js\auto.js`.
Dependency is correctly defined and transpiled to js:
```
"use strict";
sap.ui.define(["sap/ui/core/Control","./BaseChartRenderer","chart.js/auto","sap/base/Log"],
...
```
Changing from `chart.js/auto` to `chart.js` is not working.

After update to 3.3.0 this error happens:
```
info com.unitrade.crm › Running task escapeNonAsciiCharacters...
info com.unitrade.crm › Running task replaceCopyright...
info com.unitrade.crm › Running task replaceVersion...
info com.unitrade.crm › Running task ui5-tooling-transpile-task...
info com.unitrade.crm › Running task ui5-tooling-modules-task...
info com.unitrade.crm › Running task minify...
error ProjectBuilder Build failed in 2.89 s
info ProjectBuilder Executing cleanup tasks...

⚠️  Process Failed With Error

Error Message:
Resource /resources/chart.js/auto.js has no content
```
It seems like the minify tasks fails because the ui5-tooling-modules-task is not generating correct output.

**Expected behavior**
A minor feature upgrade should not break existing builds.

I've looked through the changelog and some code and cannot see why this breaks. But it is preventing me from updating and (as of now) current version is already 3.4.4.

petermuessig:
When modules have dependencies to unknown or native modules (e.g. "fs") then the tooling extension tries to include them but cannot find and open them during the build process which leads to an empty bundle.

In addition, added a proper error logging in case of bundling errors to see the stack trace for the build and the middleware. The build will stop but the middleware just prints the issue without stopping the server.

Fixes #977

### DIFF
--- a/packages/ui5-tooling-modules/lib/middleware.js
+++ b/packages/ui5-tooling-modules/lib/middleware.js
@@ -144,6 +144,9 @@ module.exports = async function ({ log, resources, options, middlewareUtil }) {
 
 			// if the resource is a bundled resource, we need to wait for it
 			const bundleInfo = whenBundled && (await whenBundled);
+			if (bundleInfo.error) {
+				log.error(bundleInfo.error);
+			}
 			const bundledResource = bundleInfo?.getEntry(moduleName);
 			if (bundledResource) {
 				resource = bundledResource;
--- a/packages/ui5-tooling-modules/lib/task.js
+++ b/packages/ui5-tooling-modules/lib/task.js
@@ -218,6 +218,10 @@ module.exports = async function ({ log, workspace, taskUtil, options }) {
 
 	// every unique dependency will be bundled (entry points will be kept, rest is chunked)
 	const bundleInfo = await getBundleInfo(Array.from(uniqueModules), config, { cwd, depPaths });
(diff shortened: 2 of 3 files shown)


petermuessig:
Test issues are related to a bug in MS OData services used for some tests

PR #995, opened 2024-05-10
fix(ui5-tooling-modules): ignore non-existing files in getResource

petermuessig:


petermuessig:
E2E issue is not related to change

PR #996, opened 2024-05-10
fix(ui5-tooling-modules): keep default export for compat with interopRequireDefault

petermuessig:


petermuessig:
E2E issue is not related

PR #998, opened 2024-05-13
fix(ui5-tooling-modules): support dynamic imports

petermuessig:


petermuessig:
E2E issue not related

PR #1004, opened 2024-05-22
feat: login url query parameter support

menof36go:
Allows specifying query parameters such as "sap-client" for onelogin. Similarly to how the simpleproxy baseUri is used as login URL when no URL is provided, the query parameters will fallback to the simpleproxy query parameters. 

Previously, you could not attach any params to the login URL at all (due to how it is constructed by appending the fiori launchpad route). In a scenario where a user is trying to proxy with a specific client, onelogin didn't care about query params and would thus try to retrieve a cookie for the wrong client. I tested this with and without simpleproxy query params and it works like a charm :)

I added some logs and some env vars for debugging and configuration convenience. I'll gladly accept any feedback, thanks in advance


petermuessig (approved):
LGTM, thx for the contribution - release follows after merge

PR #1024, opened 2024-06-27
fix(ui5-tooling-transpile): create absolute path for the replaceVersion import

drktfl:
This uses https://nodejs.org/api/url.html#urlpathtofileurlpath-options to make the path an absolute `file://` path that should be suitable for using with `import` on all platforms.

`require.resolve()` alone results in an absolute path on Windows which cannot be used.


petermuessig (approved):
LGTM

PR #1026, opened 2024-06-30
feat(ui5-tooling-modules): entry point modules must be project dependencies

petermuessig:
Starting with `ui5-tooling-modules` version `3.7.0` every entry point must be resolved from a `dependency`. Entry points used from `devDependencies` or other dependencies are not included anymore for the generation of the modules bundles. 

This ensures that the used entry points are properly declared and scan tools can identify the NPM packages for the used thirdparty modules. It also ensures that if a dependency is used the proper thirdparty modules are also available as transitive dependencies.


petermuessig:
@sz3lbi - the reason for this is that the major version is currently in sync with the major version of the UI5 tooling. We decided to do so to express the proper relationship - unfortunately, this has a negative impact for these kind of changes like this one here in which we now can't express the breaking other than the minor version change. All decision always come with pros and cons... 😢

PR #1027, opened 2024-07-03
fix: handle case when lessToCompile is provided as a string value

marcelschork:
As of now the ui5-tooling-less-task is not working when the configuration parameter `lessToCompile` is provided as a single string value.

```yaml
customTasks:
   - name: ui5-tooling-less-task
     beforeTask: minify
     configuration:
       lessToCompile: 'theme/style.less'
```

A workaround is to use a list style. 

```yaml
customTasks:
   - name: ui5-tooling-less-task
     beforeTask: minify
     configuration:
       lessToCompile: 
         - 'theme/style.less'
```
I stumbled across it when migrating an old project from a colleague to UI5 tooling v3. We used a custom less compile task supporting single values and the switch to the official UI5 tooling task made me struggling.


petermuessig:
LGTM - thanks for the fix

PR #1032, opened 2024-07-15
fix(ui5-tooling-modules): fixed some dependencies falsely assumed to be located in project root

igzThomasFrischholz:
When using ui5-tooling-modules in combination with signalr packages the dependency for "punycode" was assumed to be located in the project root and not inside node_modules subfolders. By changing the search order of file extensions in util.js this bug will not occur.

### DIFF
--- a/packages/ui5-tooling-modules/lib/util.js
+++ b/packages/ui5-tooling-modules/lib/util.js
@@ -191,7 +191,7 @@ module.exports = function (log) {
 			}
 		};
 		// try to resolve the module with different file extensions
-		for (const ext of ["", ".js", ".cjs", ".mjs"]) {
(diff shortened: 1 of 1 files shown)


petermuessig:
Hi @igzThomasFrischholz 

I also setup a small test for this but I can't see `punycode` being a dependency of `signalr`. Can you share a bit more details about your project setup? The search order is like that so that when requiring modules without their concrete path in the package they first try to resolve the NPM package with the main/module field. This would change the lookup a bit and I need to understand that first...

petermuessig:
BTW: `punycode` is also included as package in Node.js and resolves to a native module. This could be the difference here...

PR #1034, opened 2024-07-17
docs(ui5-task-zipper): fix onlyZip option

nicoschoenteich:
Shouldn't this say `false`? If set to `true`, it will only generate the zip.

### DIFF
--- a/packages/ui5-task-zipper/README.md
+++ b/packages/ui5-task-zipper/README.md
@@ -30,7 +30,7 @@ Default value: `<app-id.zip>`
 List of files to be included in the ZIP archive relative to the project root or Map of of files to be included in the ZIP archive relative to the project root and target path in the ZIP archive.
 
 - onlyZip: `true|false`
-Set this to `true` if you also want to generate the unzipped resources in the `dist` folder. Otherwise, it will only create the zipped archive.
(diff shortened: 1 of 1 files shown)


nicoschoenteich:
Updated the commit message, the check still fails though, as it wouldn't pick up the new message.

packages/ui5-task-zipper/README.md
```diff
@@ -30,7 +30,7 @@ Default value: `<app-id.zip>`
 List of files to be included in the ZIP archive relative to the project root or Map of of files to be included in the ZIP archive relative to the project root and target path in the ZIP archive.
 
 - onlyZip: `true|false`
-Set this to `true` if you also want to generate the unzipped resources in the `dist` folder. Otherwise, it will only create the zipped archive.
+Set this to `false` if you also want to generate the unzipped resources in the `dist` folder. Otherwise, it will only create the zipped archive.
```
petermuessig, line 33: I would even phrase this completely different:

Set this to `true` to omit the resources contained in the zip from the build result (typically in the `dist` folder). By default, the build result contains all resources and the zip.

petermuessig (changes requested):

petermuessig:
Can you please rebase to verify if the commit message issue still exists?

petermuessig:
THX

PR #1044, opened 2024-07-22
chore: implement new TS testing

petermuessig:


petermuessig:
@codeworrior @akudev @tobiasqueck  - FYI, this is the new way of testing in TS - the testsuites must be included as scripts and in the testsuites you need to place the annotation `/* @sapUiRequire */ which works with the `babel-plugin-transform-modules-ui5@7.6.0`

heimwege:
Hey @petermuessig 👋🏻 
Thanks for the info. I wanted to adjust my playground accordingly but then realized that everything I would need to change is part of the auto-generated code of the [preview-middleware](https://www.npmjs.com/package/@sap-ux/preview-middleware) 😏

Is `/* @sapUiRequire */` also making the `sap.ui.loader.config` shim superfluous? The [respective preview-middleware (js) template](https://github.com/SAP/open-ux-tools/blob/main/packages/preview-middleware/templates/test/qunit.js#L1) uses the shim and I want to assure consistency with what the babel-plugin-transform-modules-ui5 does.
The adjusted tests in this PR use `script` tags for the dependencies as well (e.g. for `qunit-2`) whereas the preview middleware template currently uses a `script` tag only for the `(opa|unit)Tests.qunit.js`.

petermuessig:
Hi @heimwege 

well, if the respective modules are loaded in the initial HTML page then you don't need the shims anymore. The shims are only needed to ensure that proper dependencies are loaded beforehand.

showcases/ui5-tsapp/webapp/test/integration/HelloJourney.ts
```diff
@@ -0,0 +1,48 @@
...
+	onTheMainPage.iPressTheOkButtonInTheDialog();
+
+	// Assertions
+	onTheMainPage.iShouldNotSeeTheHelloDialog();
+
+	// Cleanup
+	onTheMainPage.iTeardownMyApp();
+});
+
+opaTest("Should close the Hello dialog", function (Given, When, Then) {
```
akudev, line 32: ummm... didn't we get rid of Given/When/Then? It's also not used.

petermuessig: Seems I missed to remove it at this place... THX for the catch

akudev:
@petermuessig  When this is the new way, what is the plan for e.g.
https://github.com/SAP-samples/ui5-typescript-tutorial/tree/main/exercises/ex6/com.myorg.myapp/webapp/test/unit
and
https://github.com/SAP-samples/ui5-typescript-helloworld/tree/testing/webapp/test/unit
?

petermuessig:
I have it on my list to update the helloworld, the tutorial and the walkthrough

PR #1046, opened 2024-07-22
feat(ui5-task-zipper): optional  parameter to turn absolute data sour…

nicoschoenteich:
…ce paths into relative ones

I stumpled across this while working on the [generator-ui5-project](/ui5-community/generator-ui5-project), where we want to turn absolute data sources paths into relative ones, to be able to deploy them to SAP Build Work Zone, standard edition. Currently the [open-ux-tools](https://github.com/SAP/open-ux-tools) (which both the generator-ui5-project and the Fiori Tools Application Generator rely on) only add absolute paths to the manifest, which is why this transformation is necessary.

I thought it makes sense to implement this here in a reusable manner instead of building a custom script for it.

This has been mentioned/requested before, see here: https://github.com/SAP/ui5-tooling/issues/625

### DIFF
--- a/packages/ui5-task-zipper/lib/zipper.js
+++ b/packages/ui5-task-zipper/lib/zipper.js
@@ -1,7 +1,5 @@
 const path = require("path");
 const yazl = require("yazl");
-const { Buffer } = require("node:buffer");
-const fs = require("fs");
 
 /**
  * Determines the project name from the given resource collection.
@@ -25,7 +23,6 @@ const determineProjectName = (collection) => {
 
 /**
   * Turn absolute data source paths in the manifest.json into relative paths
-  * Only works if manifest.json is located in the `webapp/` directory
   *
   * @param {Buffer} buffer Buffer of the manifest.json file
   * @param {object} zip ZipFile instance
--- a/packages/ui5-task-zipper/lib/zipper.js
+++ b/packages/ui5-task-zipper/lib/zipper.js
@@ -1,5 +1,7 @@
 const path = require("path");
 const yazl = require("yazl");
+const { Buffer } = require("node:buffer");
+const fs = require("fs");
 
 /**
  * Determines the project name from the given resource collection.
@@ -22,6 +24,25 @@ const determineProjectName = (collection) => {
 };
 
 /**
+  * Turn absolute data source paths in the manifest.json into relative paths
+  * Only works if manifest.json is located in the `webapp/` directory
+  *
+  * @param {Buffer} buffer Buffer of the manifest.json file
+  * @param {object} zip ZipFile instance
+  * returns {void}
+*/
+const absoluteToRelativePaths = (buffer, zip) => {
+	const manifest = JSON.parse(buffer.toString("utf-8"));
+	if (!manifest["sap.app"]["dataSources"]) return;
+	for (const [dataSourceName, dataSource] of Object.entries(manifest["sap.app"]["dataSources"])) {
+		if (dataSource.uri?.substring(0, 1) === "/") {
+			manifest["sap.app"]["dataSources"][dataSourceName].uri = dataSource.uri.substring(1);
+		}
+	}
+	zip.addBuffer(Buffer.from(JSON.stringify(manifest, null, 4), "utf-8"), "manifest.json");
+}
+
+/**
  * Zips the application content of the output folder
  *
  * @param {object} parameters Parameters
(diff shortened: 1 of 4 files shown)


packages/ui5-task-zipper/lib/zipper.js
```diff
@@ -1,5 +1,7 @@
 const path = require("path");
 const yazl = require("yazl");
+const { Buffer } = require("node:buffer");
```
petermuessig, line 3: either all built-in modules with `node:` or none

nicoschoenteich:
Thanks for the review 😊 Ready to merge now?

packages/ui5-task-zipper/lib/zipper.js
```diff
@@ -95,7 +116,11 @@ module.exports = async function ({ log, workspace, dependencies, options, taskUt
 					zipEntries.push(resourcePath);
 					return resource.getBuffer().then((buffer) => {
 						isDebug && log.info(`Adding ${resource.getPath()} to archive.`);
-						zip.addBuffer(buffer, resourcePath); // Replace first forward slash at the start of the path
+						if (relativePaths && resourcePath.includes("manifest.json")) {
```
petermuessig, line 119: Sorry, missed that one - includes is wrong here - as it should be matching, right? But maybe you want to handle all included manifest.json files - but then you need to pass also the `resourcePath` to the method `absoluteToRelativePaths` to put the file into the proper location again? Otherwise all manifest.json files e.g. manifest.json, reuse-component/manifest.json will overwrite the *manifest.json* in the root. But you could also return the modified buffer from the function and then move zip.addBuffer outside of the else branch so that it runs also for the modification case.

```js
if (relativePaths && resourcePath.includes("manifest.json")) {
  buffer = absoluteToRelativePaths(buffer);
}
zip.addBuffer(buffer, resourcePath); // Replace first forward slash at the start of the path
```

packages/ui5-task-zipper/test/zipper.test.js
```diff
@@ -11,9 +11,10 @@ const test = require("ava");
  *
  * @param {string} zip path to a zip file
  * @param {string} needle path/file to find in the zip
- * @returns {Promise<true|false>} true when needle is found in zip, false otherwise
+ * @param {boolean} readFileContent whether to read the file content
+ * @returns {Promise<true|string|false>} true when needle is found in zip, string if file content could be read, false otherwise
  */
-async function promisifiedNeedleInHaystack(zip, needle) {
+async function promisifiedNeedleInHaystack(zip, needle, readFileContent = false) {
```
petermuessig, line 17: Setting the default to `false` is optional as normally the unset parameter is `undefined` which is a falsy value

petermuessig (changes requested):
Sorry, one more - I didn't see this in my check yesterday - but then we can merge!

packages/ui5-task-zipper/lib/zipper.js
nicoschoenteich: Good catch. Let's only modify the `manifest.json`. I will change it to `resourcePath === "manifest.json"`.

packages/ui5-task-zipper/lib/zipper.js
```diff
@@ -21,6 +21,24 @@ const determineProjectName = (collection) => {
...
+  *
+  * @param {Buffer} buffer Buffer of the manifest.json file
+  * @param {object} zip ZipFile instance
+  * returns {void}
+*/
+const absoluteToRelativePaths = (buffer, zip) => {
+	const manifest = JSON.parse(buffer.toString("utf-8"));
+	if (!manifest["sap.app"]["dataSources"]) return;
+	for (const [dataSourceName, dataSource] of Object.entries(manifest["sap.app"]["dataSources"])) {
+		if (dataSource.uri?.substring(0, 1) === "/") {
```
flovogt, line 35: usage of [`charAt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt) would be a little bit shorter

packages/ui5-task-zipper/test/zipper.test.js
nicoschoenteich: Correct, I personally like setting defaults as it improves readability.

packages/ui5-task-zipper/lib/zipper.js
nicoschoenteich: Yep

vc10812:
@nicoschoenteich @petermuessig 
The release action was skipped. When will the latest version be available from npm?

nicoschoenteich:
@petermuessig usually does the releases. Are we good to release?

petermuessig:
Sorry, I messed-up the release - should be available now

@vc10812 @nicoschoenteich

nicoschoenteich:
Hi @norbip, that is a good point. I will look into that.

nicoschoenteich:
Hi @norbip, I opened a [PR](https://github.com/ui5-community/ui5-ecosystem-showcase/pull/1152) with the fix.

PR #1053, opened 2024-07-31
feat(ui5-middleware-onelogin): allow changing login subdirectory & retry certificate sign-on

menof36go:
Currently, onelogin is hard-coded to go to the Fiori Launchpad to retrieve the cookie. This is a really strict limitation, especially since we could technically use any page that will ask for a login. For example, I have a system that only exposes OData services (MII java stack), but does not host a FLP and thus couldn't use onelogin with this limitation. I made the subdirectory configurable with a default value corresponding to the currently hardcoded value (Thus this change should be non-breaking).

I also noticed that the certificate login is sometimes a little flaky and will just randomly throw me on a login page. In this case, reloading the login page resolves the issue. To solve this issue programmatically, onelogin will now look at the login page after it has been loaded and if it looks like a login page, refresh it.

I'll gladly accept any feedback, thanks in advance


petermuessig:
THX for your PR, LGTM

petermuessig (approved):

petermuessig:
Released with:
 - ui5-middleware-onelogin: 3.5.0

PR #1055, opened 2024-08-08
fix(ui5-task-zipper): make sure manifest gets zipped if no data sourc…

nicoschoenteich:
…es exist


petermuessig:
Released with:
 - ui5-task-zipper: 3.2.1

PR #1060, opened 2024-08-13
fix(ui5-tooling-modules): perf-opt for module type detection

petermuessig:


petermuessig:
@hschaefer123 - one more to come 😉

PR #1096, opened 2024-10-17
fix(ui5-tooling-modules): named aggregations can also be marked as default aggregation

Thodd:


packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -143,18 +144,24 @@ class RegistryEntry {
...
 		const multiple = !!arrayTypeMatch;
-		parsedType = arrayTypeMatch?.[1] || parsedType;
 
-		// complex types have a reference to other things, e.g. enums
-		if (typeInfo?.references) {
+		// [Complex types]:
+		// we have a reference to other things -> enums, interfaces, classes
+		if (typeInfo?.references?.length > 0) {
+			// Since the UI5 runtime only allows for 1 single type per property/aggregation, we take the first reference
+			parsedType = typeInfo.references[0].name;
```
petermuessig, line 159: Ah, ok - this is interesting because the Luigi guys also used multiple types. So, the first type wins and should be then the most generic type - an alternative would be to use `object` in such cases.

petermuessig (approved):

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
Thodd: Yes, the runtime only evaluates one single type.
You can document more than one of course. To support multiple ones at runtime, you could also use `any` e.g. if you want to accept things like `object | string`.

PR #1101, opened 2024-10-18
fix(ui5-tooling-modules): properly support keepDynamicImports

petermuessig:


petermuessig:
@JohannesDoberer @hardl

this change will now ensure that the dynamic imports are kept by default when using variables.

PR #1108, opened 2024-10-24
feat(ui5-tooling-modules): support special UI5 settings for web components

Thodd:
#### TODO:
* [x] Import of LabelEnablement mixin for the webc that implement `sap.ui.core.Label`
* [x] Introduce Mapping for
  * [x] enabled/diabled
  * [x] textDirection/dir

#### Next Steps:
* Test how custom data (or a variation) could be used to handle native properties (see below, _Open special properties/events_)
* Investigate how to implement `sap.ui.core.ISemanticFormContent` for the Multi Input?
  * Is this even needed?
  * What does the semantic form content do in UI5 anyway?

#### Questions:
* ~~Should the `for` property of `@ui5/webcomponents/Label` be renamed to `labelFor` like in the original retrofits?~~
  * **Answer**: No we will not align with the deprecated retrofit components. The custom elements metadata will be leading.
* ~~Is `InputElementsFormSupport` still needed for the latest WebC versions?~~
  * ~~Seems to be replaced by an actual interface...~~
  * **Answer**: No it is not needed anymore. The current web components implementation uses a newer internal interface approach, refer to [OpenUI5Enablement.ts](https://github.com/SAP/ui5-webcomponents/blob/f3fd7c0139ca4672138917b26d5b0a6ec902f761/packages/base/src/features/OpenUI5Enablement.ts)

### DIFF
--- a/packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
+++ b/packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
@@ -375,18 +375,6 @@ class RegistryEntry {
 			};
 		}
 
-		// TODO: Discuss if this is still needed.
-		//       Aren't we wrongfully introducing aggregations to controls that shouldn't have any.
-		//       e.g. the "@ui5/webcomponents.Switch" does not have a default slot and thus shouldn't have "content" aggregation.
-		// mandatory default aggregation, named "content" in UI5
-		// if (!ui5metadata.defaultAggregation) {
-		// 	ui5metadata.aggregations["content"] ??= {
-		// 		type: "sap.ui.core.Control",
-		// 		multiple: true,
-		// 	};
-		// 	ui5metadata.defaultAggregation = "content";
-		// }
-
 		// cssProperties: [ "width", "height", "display" ]
 		if (!ui5metadata.properties["width"]) {
 			ui5metadata.properties["width"] = {
--- a/packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
+++ b/packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
@@ -228,28 +228,69 @@ class RegistryEntry {
 		}
 	}
 
+	/**
+	 * Checks the given property definition object for a special UI5 mapping.
+	 * The specific formatter are implemented in "sap.ui.core.webc.WebComponent".
+	 *
+	 * @param {object} propDef the property definition object from the custom elements metadata
+	 * @param {object} ui5metadata the UI5 metadata object
+	 * @returns {boolean} whether the property needs a special mapping
+	 */
+	#checkForSpecialMapping(propDef, ui5metadata) {
+		if (propDef.name === "accessibleNameRef") {
+			ui5metadata.associations["ariaLabelledBy"] = {
+				type: "sap.ui.core.Control",
+				multiple: true,
+				mapping: {
+					type: "property",
+					to: "accessibleNameRef",
+					formatter: "_getAriaLabelledByForRendering",
+				},
+			};
+			return true;
+		} else if (propDef.name === "disabled") {
+			// "disabled" maps to "enabled" in UI5
+			ui5metadata.properties["enabled"] = {
+				type: "boolean",
+				defaultValue: "true",
+				mapping: {
+					type: "property",
+					to: "disabled",
+					formatter: "_mapEnabled",
+				},
+			};
+			return true;
+		} else if (propDef.name === "textDirection") {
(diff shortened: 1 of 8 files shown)


packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -365,14 +375,17 @@ class RegistryEntry {
...
 
-		// mandatory default aggregation
-		if (!ui5metadata.defaultAggregation) {
-			ui5metadata.aggregations["default"] = {
-				type: "sap.ui.core.Control",
-				multiple: true,
-			};
-			ui5metadata.defaultAggregation = "default";
-		}
+		// TODO: Discuss if this is still needed.
```
Thodd, line 378: See TODO

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -390,6 +403,45 @@ class RegistryEntry {
...
+	#patchUI5Specifics(ui5metadata) {
+		const { tag } = ui5metadata;
+
+		// The label has a couple of specifics that are not fully reflected in the custom elements.
+		// This code harmonizes the Label with the original retrofit control wrapper.
+		// TODO: How to generalize this?
+		//       Is harmonizing with the retrofit library really correct?
+		if (tag === "ui5-label") {
+			// the ui5-label has as default slot, but no aggregations on the retrofit layer...
+			ui5metadata.aggregations = [];
```
Thodd, line 423: Everything except adding the interface in line :438, might not be needed if we decide to keep the naming as `for`.

Only difference is that `for` is a property, where as I think it should be an association in UI5.
Seems like a special case where the type is not `HTMLElement` (as it is with other associations). The `for` property is of type `string | undefined`. According to the documentation, the attribute value must be an ID.

I guess that should then be an association in UI5... 🤔

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -390,6 +403,45 @@ class RegistryEntry {
...
+			};
+			delete ui5metadata.properties["for"];
+
+			// Any "Label" control needs a special UI5-only interface
+			// Additionally, all controls implementing this interface must apply the "sap/ui/core/LabelEnablement" mixin on their class
+			// refer to "../templates/WrapperControl.hbs"
+			ui5metadata.interfaces.push("sap.ui.core.Label");
+		} else if (tag === "ui5-multi-input") {
+			// TODO: Multi Input needs to implement the functions defined in "sap.ui.core.ISemanticFormContent"...
+			ui5metadata.interfaces.push("sap.ui.core.ISemanticFormContent");
```
Thodd, line 441: I don't know yet what this actual does... need to ask Sebastian about it.

showcases/ui5-tsapp-webc/webapp/view/Main.view.xml
```diff
@@ -73,7 +73,7 @@
 
           <webc:Popover id="popover1" opener="popoverOpenerButton1" headerText="Newsletter subscription" placement="Bottom">
             <c:HTMLElement tag="div" class="popover-content">
-              <webc:Label for="emailInput" required="true" showColon="true" text="Email" />
+              <webc:Label labelFor="emailInput" required="true" showColon="true" text="Email" />
```
Thodd, line 76: can be reverted if we keep the metadata defined name `for`.

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
Thodd: So turns out we do need this harmonization for the Label control :(
The `sap/ui/core/LabelEnablement` expects an association named `labelFor`.

Thodd: Discussed and approved by PM.
Later down the line we can introduce a specific metadata setting.

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
Thodd: I'd keep this for now, as the interface itself is not wrong. But we need to clarify what the implications are:
`Form` vs `SemanticForm`?

petermuessig: Ok, anyhow we go ahead an merge this change since it brings some new valuable features to align with the older retrofit by adding the enabled and textDirection mapping and adding the LabelSupport.

PR #1111, opened 2024-10-30
fix(ui5-tsapp-simple): mark SimpleControl renderer as static

Thodd:
fix(ui5-tsapp-webc): mark HTMLElement renderer as static


codeworrior (approved):
LGTM, the `renderer` is not instance specific in UI5.

petermuessig (approved):

PR #1112, opened 2024-10-30
feat(ui5-tooling-modules): support including Assets for Web Components

petermuessig:


petermuessig:
@Thodd @vladitasev

We need to re-discuss the `OpenUI5Support` in general. When including the Assets modules per package it comes most probably too late and then the initial theme can't be applied correctly. This also applies to the languages and all other dynamic imports.

Thodd:
> We need to re-discuss the `OpenUI5Support` in general. When including the Assets modules per package it comes most probably too late and then the initial theme can't be applied correctly. This also applies to the languages and all other dynamic imports.

Hm I will look into this later, not sure why this would be too late 🤔
If I see this correctly, the retrofit wrappers load the OpenUI5Support from the `sap/ui/webc/common/library.js`.
That should be the earliest point in time, correct @vladitasev?

petermuessig:
It's the loading order:

webcomponents-fiori > webcomponents > webcomponents-base > OpenUI5Support
webcomponents-fiori > Assets

It's the nature of the import order we have - but we can't change that easily except of moving the OpenUI5Support initialization to a later point in time...

Today it works by manually importing the Assets in the Component.js which is also OK but it would require documentation how to include additional styles and exclude the dynamics from the preload otherwise the preload is ~10MB unzipped! 😓

PR #1116, opened 2024-11-01
fix(ui5-tooling-modules): include LabelEnablement to list of external deps

petermuessig:


petermuessig:
@Thodd : when adding external deps (UI5 runtime resources) to the templates as dependencies, you need to mark them as external to avoid a warning in the console and an unnecessary lookup.

PR #1122, opened 2024-11-13
fix(ui5-tooling-modules): apply sap/ui/core/EnabledPropagator mixin

Thodd:
When the `enabled` (resp. `disabled`) property exists, we need to also apply the `sap/ui/core/EnabledPropagator` mixin to the control wrapper.
This way the `enabled` state is propagated along the control tree.


petermuessig (dismissed):
LGTM

petermuessig:
Unfortunately, you need to update the tests with: `pnpm --filter ui5-tooling-modules test:snapshots`

petermuessig:
About the wdi5 issue - hmm, don't know yet - maybe a one time issue...

PR #1140, opened 2025-01-02
fix(ui5-tooling-modules): ignore scoping for non-UI5Elements

petermuessig:


petermuessig:
@pskelin - the issue related to the scoping is that some Web Components do not provide the proper metadata to be a UI5Element which is an indicator to support scoping! (related to Account Cockpit issue before the holidays)

With this change, I'll skip non-UI5Elements to use scoping.

PR #1164, opened 2025-02-06
Do not merge/js doc

petermuessig:


petermuessig:
@thodd :

this morning I discussed with @codeworrior to not patch the api.json and rather create a `api-mixin[-xxx].json` next to the api.json which can be collected by the jsdoc generation to patch the apis.json with the provided content. Then we can prepare this during the regular build in our existing task and just drop it into the test-resources/%namespace%/designtime location.

petermuessig:
Closed in favor of #1171

PR #1177, opened 2025-03-04
fix(ui5-tooling-modules): ensure full rebuild after change of modules

petermuessig:
The mechanism in the rollup plugin to generate the package wrappers relies on the lookup of the WebComponentsRegistry. When the packages and classes are known to the registry the packages will not be emitted anymore. In any case we nevertheless need to rebuild the web components metadata in case of changes to them. So the clear function now triggers a complete reload.


petermuessig:
@Thodd I need to reset the `WebComponentRegistry? for each build iteration to ensure that the metadata is properly loaded and to guarantee that all relevant modules are properly emitted. In future we may think about improvements to this.

PR #1178, opened 2025-03-05
chore: improve Luigi demo

petermuessig:
ℹ️  @hardl @JohannesDoberer

Interesting ist the following change: https://github.com/ui5-community/ui5-ecosystem-showcase/pull/1178/files#diff-bd26999978b6bb58b9de1b1a32ae9738c5ee5cabe422403a147c6085bc327ca0


petermuessig:
@Thodd with that change, I re-export the exports from the original NPM package - e.g. `@luigi-project/container` exports already some modules and with this change I ensure that a re-export takes place. Therefore, I enhanced the module generation for the packages. 

BTW: I noticed also that packages already exports their enums properly - so no action item here! @codeworrior 

@hardl @JohannesDoberer - with this change now, the import of `LuigiEvents` works out of the box and it matches the type definitions.

PR #1191, opened 2025-03-25
feat(ui5-tooling-modules): support special UI5 settings for ui5-shell…

adrian-bobev:
…bar interfaces


petermuessig:
@adrian-bobev : one question - is this UI5 runtime version related and we should enable this only starting with a specific version?

petermuessig:
@adrian-bobev : before I can merge it, the tests need to be updated - can you please run the following command and upload the changed fixtures:

```sh
pnpm --filter ui5-tooling-modules test:fixtures
```

petermuessig:
Fix is released

Issue #1028, opened 2024-07-04
[cds-plugin-ui5] Option to only include / exclude apps during server start up


jbibal-pdc:
**Is your feature request related to a problem? Please describe.**
When using the plugin with a big project, with multiple Fiori / UI5 apps, running the CAP / CDS server becomes very slow as it needs to spawn a server for each of the apps, when there are times that we only want to test and work on a particular app.

**Describe the solution you'd like**
Probably a config wherein we can specify which apps to **include** or maybe **exclude** when starting the server.

**Describe alternatives you've considered**
Break down the apps into different projects, which is a hassle since the apps share the same domain models.

**Additional context**
![image](https://github.com/ui5-community/ui5-ecosystem-showcase/assets/105326887/4bd26e84-4792-4744-981c-db44fcf3e0f8)

petermuessig:
@jbibal-pdc  Sorry for my late reply - the idea is great - I'm just thinking how to do it - would be a `.dotenv` file or environmental variables OK to control the apps to be started as a comma separated list with the NPM package names? WDYT?

jbibal-pdc:
Hi @petermuessig, thanks for replying and considering my suggestion. Yes, maybe a `.dotenv` file or an environment variable. 

Personally, I was also trying to think of a good way to handle it as part of this feature request but couldn't think of a cleaner implementation.

While awaiting a response though, I did play around with it on my local and added support for a custom config in `package.json` in the form of `pkgJson.cds?.["cds-plugin-ui5"]?.includeModulesOnly`.

Like if this setting is `true`, then only those with a specified config under the `pkgJson.cds?.["cds-plugin-ui5"]?.modules` is included during testing, specifically, during the time the `localApps` array is populated.

Something like:

```
let includeModulesOnly = pkgJson.cds?.["cds-plugin-ui5"]?.includeModulesOnly || false;

// Then during the LOCAL apps iteration
if (!skipLocalApps) {
    const appDir = path.join(cwd, cds?.env?.folders?.app || "app");
    if (fs.existsSync(appDir)) {
        // is the UI5 app directly in the app directory?
        if (!fs.existsSync(path.join(appDir, determineUI5Yaml(appDir)))) {
            // lookup all dirs inside the root app directory for
            // being either a local app or a UI5 application
            fs.readdirSync(appDir, {
                    withFileTypes: true
                })
                .filter((f) => f.isDirectory())
                .forEach((d) => {

[...]

PR #1193, opened 2025-03-26
feat(cds-plugin-ui5): add lazy module loading (#WIP)

stockbal:
Hi @petermuessig,

hopefully you didn't have time yet to build a solution for #1028 😅. I was playing around with the lazy middleware loading approach, that you described as possible solution.
It is not finished yet, I just wanted to get your opinion if this goes in the direction you imagined. Generally the lazy loading works. The routers are mounted for the module paths at startup - which I guess we can't get rid off?? -  but the middlewares are only loaded once the first HTTP call goes to a certain module path.

Possible fix for #1028.

### DIFF
--- a/packages/cds-plugin-ui5/cds-plugin.js
+++ b/packages/cds-plugin-ui5/cds-plugin.js
@@ -172,6 +172,8 @@ if (!skip) {
 				// log the version of the cds-plugin-ui5
 				logVersion();
 
+				const isLazyLoadingEnabled = process.env["CDS_PLUGIN_UI5_LAZY_LOADING"] === "true";
+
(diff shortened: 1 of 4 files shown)


petermuessig:
Hi @stockbal ,

indeed, the lazy loading is a pretty good idea. I need to verify this whether it works in all scenarios. I didn't think about this issue so far. Good that you remind me on that.

Cheers
Peter

petermuessig:
Finally @stockbal I have the chance to check your draft. I'll tweak it a bit more so that all HTML pages are displayed properly. But overall, small and nice enhancement!

PR #1210, opened 2025-05-07
feat(ui5-tooling-modules): jsdoc generation support

### LINKED ISSUE #1208: [ui5-tooling-modules] Third-party dependency is skipped though it is listed in package.json as dependency
**Describe the bug**
I have a UI5 application with a third-party dependency and a UI5 library. In our current setup, we require the library and the third-party dependency `ajv` as a dependency in the `package.json`.
As soon as I integrated the library, the deployed application stopped working as the third-party dependency was not longer included in the build output - the following warning was shown in the logs of the build process:
`Skipping dependency "ajv" as it is not part of the dependencies in package.json!``

The `dependencies` in the `package.json` of the application looks like the following:
```json
"dependencies": {
    "<namespace>.reuse": "file:../reuse-lib"
    "ajv": "^8.17.1",
    "ajv-formats": "^3.0.1",
    "ajv-i18n": "^4.2.0"
 }
```
Currently, the library itself does not have any `dependencies`.

Additionally, there was an issue in the past which required us to add `ajv` as an additional dependency in the `ui5.yaml` of the application itself:
```yaml
configuration:
        addToNamespace: true
        additionalDependencies: ["ajv"]
```

During the dependency resolution process in `ui5-tooling-modules`, the module looks for `ajv` in the `node_modules` directory of the library. This happens because `ajv` is also a dependency of `eslint`, which is used as a `devDependency` within the library. Consequently, `ajv` becomes a transitive dependency of the library through `eslint`, and the resolution process identifies it. As a result, `ajv` is added to `depRoots`, associating it with the library's path.
```js
if (depRoot && depRoots.indexOf(depRoot) === -1) {
  depRoots.push(depRoot);
  depRoots.push(...findDependencies({ cwd: depRoot, depPaths, linkedOnly, additionalDeps }, knownDeps));
}
```

Later, `ajv` is also found as a dependency in the `node_modules` of the application. As this is already a known dependency, the path of the dependency is not added to the dependency roots:
```js
if (knownDeps.indexOf(npmPackage) !== -1) {
  continue;
}
```

This leads to `allowedDep` being set to false during dependency resolution in the function `addDep`:
```js
let allowedDep = existsDep && (!dependencyRoots || dependencyRoots.some((root) => depPath.startsWith(root)));
```

Therefore, the modules logs the warning above. For my use case, the issue is solved by removing the flag `additionalDependencies`. As I think that the same dependency could be required by a library and a application in a similar setup, I raise this issue so others can have a better experience setting this up.

Imo, it is necessary to push the third-party dependency to the `dependencyRoots` for all found occurrences, not only when it is an unkown dependency (not included in the `knownDep` arrray). I am sure that I only have touched the surface of the dependency resolution mechanism and there are much more side effects to consider.

**To Reproduce**
If required, I can provide a sample internally.

**Expected behavior**
The dependency resolution should recognize the dependecy correctly so that the build artefact is working and able to load the dependency during runtime.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: 3.20.2 (of ui5-tooling-modules)

**Additional context**
N/A

petermuessig:
Fixes #1208 (as a side effect)


petermuessig:
Force push to update PR with the latest adoptions - to run the JSDoc generation for Seamless WebC you can run the following command: 

`pnpm --filter ui5-tslib jsdoc`

petermuessig:
@Thodd @H4ze - change is ready and validated with usage scenarios. Change works and can be submitted if you are OK with it...

packages/ui5-tooling-modules/lib/templates/jsdoc/ClassHeader.hbs
```diff
@@ -5,5 +5,4 @@
  * @constructor
  * @public
  * @alias {{{alias}}}
- * @ui5-module-override {{{aliasSlashed}}}
```
petermuessig, line 8: Do we still need this?

PR #1211, opened 2025-05-08
fix(dev-approuter): switch to custom logger to avoid loading cds.env …

nicoschoenteich:
…too early

### DIFF
--- a/packages/dev-approuter/package.json
+++ b/packages/dev-approuter/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dev-approuter",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "description": "A dev time wrapper for the SAP Application Router that can serve UI5 and CDS modules added as dependencies.",
   "author": "Nico Schoenteich <nicolai.schoenteich@sap.com> (https://github.com/nicoschoenteich)",
   "license": "Apache-2.0",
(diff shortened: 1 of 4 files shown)


packages/dev-approuter/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dev-approuter",
-  "version": "0.5.1",
+  "version": "0.5.2",
```
petermuessig, line 3: @nicoschoenteich : you do not need to increase the version manually, this will happen with the release build!

packages/dev-approuter/package.json
```diff
@@ -16,13 +16,5 @@
     "cds-plugin-ui5": "workspace:^",
     "express": "^4.21.2",
     "ui5-middleware-cap": "workspace:^"
-  },
-  "peerDependencies": {
```
petermuessig, line 20: There are no dependencies to @sap/cds anymore?

petermuessig (changes requested):

packages/dev-approuter/package.json
nicoschoenteich: Correct, only as a second level dependency via the `ui5-middleware-cap`.

packages/dev-approuter/package.json
nicoschoenteich: Ok

petermuessig (approved):

petermuessig:
I'll release it a bit later - need to prepare a change for tooling-modules right now which I release together with it.

PR #1215, opened 2025-05-16
fix: [cds-plugin-ui5] check that it's the `cds` script

chgeo:
The version check for `@sap/cds-dk` is currently wrong if another script runs like `cds-typer`.

Reproduce with e.g.
```sh
npx -p @cap-js/cds-typer cds-typer "*" 
```
and you get 
```
[cds-plugin-ui5] [WARN] The cds build task requires @sap/cds-dk version >= 7.5.0! Skipping execution as your @sap/cds-dk version is too old...
```

Fix is to check if it's really the `cds` script that runs.


chgeo:
@petermuessig please review

(PR diff is best viewed with whitespace diff off).

petermuessig (approved):
LGTM - I'll correct the commit message before the merge

PR #1216, opened 2025-05-19
feat(cds-plugin-ui5): add lazy module loading

### LINKED ISSUE #1028: [cds-plugin-ui5] Option to only include / exclude apps during server start up

petermuessig:
This feature introduces lazy loading for UI5 applications and closes the POC of @stockbal (BTW: big thanks for that! nice idea!)

Closes #1193
Fixes #1028

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "dev-simple:ts": "pnpm --filter ui5-tsapp-simple dev",
     "dev-webc:ts": "pnpm --filter ui5-tsapp-webc dev",
     "dev-cds": "pnpm --filter cds-bookshop watch",
+    "dev-cds:lazy": "pnpm --filter cds-bookshop watch:lazy",
     "dev-bookshop": "pnpm --filter ui5-bookshop-viewer dev",
     "dev-approuter": "pnpm --filter approuter dev",
     "start": "pnpm --filter ui5-app start",
--- a/packages/cds-plugin-ui5/README.md
+++ b/packages/cds-plugin-ui5/README.md
@@ -164,6 +164,8 @@ module.exports = async ({ log, resources, options }) => {
 
 The returned app pages will be added to the welcome page within the respective mount path.
 
+> :warning: The app pages cannot be retrieved and injected when using the lazy loading option.
+
 ## Hints
 
 This section includes hints for the usage of the `cds-plugin-ui5` with other tools.
--- a/packages/cds-plugin-ui5/README.md
+++ b/packages/cds-plugin-ui5/README.md
@@ -75,6 +75,15 @@ The configuration can also be injected with the environment variable `CDS_PLUGIN
 CDS_PLUGIN_UI5_MODULES="{ \"ui5-bookshop\": { \"mountPath\": \"/the-bookshop\" } }" cds-serve
 ```
 
+#### Lazy Loading
+
+The plugin supports lazy loading of UI5 applications which means that the UI5 middlewares will not be applied by default. The first time a UI5 application will be accessed in the CDS server triggers the load and apply of the UI5 middlewares. This feature is not active by default and needs to be activated with the environment variable `CDS_PLUGIN_UI5_LAZY_LOADING`.
+
+```sh
+# enable lazy loading for UI5 applications
+CDS_PLUGIN_UI5_LAZY_LOADING=true cds watch
+```
+
 #### Logger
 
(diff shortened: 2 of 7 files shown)


petermuessig:
@stockbal - WDYT?

petermuessig:
Yes, indeed - it's a bit a trade-off between using the UI5 tooling APIs, reading the `ui5.yaml` to determine the application pages. I need to see whether I can even improve that part a bit more. Thanks for the fast reply and feedback! 👍

petermuessig:
Some interesting performance measurements:

```
# Launch CDS server without blocking applyUi5Middleware function:
[cds] [INFO] server listening on { url: 'http://localhost:4004' }
[cds] - launched at 5/20/2025, 9:39:08 AM, version: 8.1.1, in: 883.272ms
[cds] [INFO] [ terminate with ^C ]

# Numbers from measuring the internals of the applyUi5Middleware function:
import took 482ms
Graph took 628ms
Root project took 0ms
Pages took 4ms
applyUI5Middleware took 1112ms

import took 479ms
Graph took 630ms
Root project took 0ms
Pages took 3ms
applyUI5Middleware took 1112ms

Graph took 633ms
Root project took 0ms
import took 479ms
Pages took 1ms
applyUI5Middleware took 1117ms
```

As all of those 3 projects are running in parallel the import `await` is shared across all 3 and the graph initializiation is always around 500ms. Maybe also there most of the time is lost due to the initialization of the required modules. Seems that this is the most expensive piece.

Anyhow => I finally decoupled the `applyUi5Middleware` from the initialization phase and made the links modifiable lazily. So, each application can inject it's pages once initialized and the initialization doesn't block the startup of the CDS server anymore. You can see this in the log above, the CDS server is there before applyUi5Middleware completes. In general, the times above from applyUi5Middleware is anyhow shared so it is not 3secs in total to prepare the apps, it's delaying the availability for all 3 of them by 1sec. And the middlewares are started anyhow later on first request!

stockbal:
Just tested it with a project containing 28 apps (😅). It took 8s for cds to start.
Nicccccccce 🎉🪅.

petermuessig:
I'm going to release it.

marianfoo:
Amazing! That always bugged me to wait for all the apps.
Thank you all!

hschaefer123:
great performance update. thx

petermuessig:
@pwasem - could be related to the lazy startup which has a marked property to know if it is already started. maybe this causes a side effect - I have to check... The issue appears by simple change in a cds file and then a reload of cds server?

pwasem:
> @pwasem - could be related to the lazy startup which has a marked property to know if it is already started. maybe this causes a side effect

In `.env`  have already set:

```shell
# https://github.com/ui5-community/ui5-ecosystem-showcase/pull/1216/files
CDS_PLUGIN_UI5_LAZY_LOADING=false
```

> The issue appears by simple change in a cds file and then a reload of cds server
Yes, the reload happens to early and the app has not been served at this point of time

stockbal:
@petermuessig  I noticed the same thing. I was so happy about the improvement of the startup time that I did do any deeper testing 😅. Now as I switched from cds-ts to cds-tsx the issue became even more noticable, as the cds startup reduced to 1.5s for my app with >25 UI5 modules, but it actually takes an additional 12s for the UI5 modules to be fully mounted. Which has the effect that the cds startup page at `localhost:4004` does not list any UI5 app links initially, and also results in the 404 issues after a cds restart as mentioned by @pwasem.

The root cause is the `applyUI5Middleware` call in https://github.com/ui5-community/ui5-ecosystem-showcase/blob/a0743106a7c31f7a33fe69d34035402fb0a69c7b/packages/cds-plugin-ui5/cds-plugin.js#L191
Because it is not awaited, the mounting process itself is not done yet, when the cds server bootstrapping is complete, so the reload-request for a UI5 app module can not by handled.

PR #1217, opened 2025-05-20
fix(ui5-tooling-modules): add @implements tag to wrapper JSDoc

codeworrior:
The @implements tag in the class header documentation for a wrapper class should list the JSDoc names of the implemented interfaces. In contrast, runtime metadata contains the runtime metadata names of the interfaces.

### DIFF
--- a/packages/ui5-tooling-modules/lib/templates/jsdoc/ClassHeader.hbs
+++ b/packages/ui5-tooling-modules/lib/templates/jsdoc/ClassHeader.hbs
@@ -1,9 +1,7 @@
 /**
  * @class
- * {{{description}}}
-{{#if interfacesSlashed}}
- * @implements {{{interfacesSlashed}}}
-{{/if}}
+ * {{{description}}}{{#if interfacesSlashed}}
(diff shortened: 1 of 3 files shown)


packages/ui5-tooling-modules/lib/templates/jsdoc/ClassHeader.hbs
```diff
@@ -1,6 +1,10 @@
 /**
  * @class
  * {{{description}}}
+{{#if interfacesSlashed}}
```
petermuessig, line 4: to avoid empty lines in the output, please move the conditional handlebars tags into the previous line, e.g.:

```
/**
 * @class
 * {{{description}}}{{#if interfacesSlashed}}
 * @implements {{{interfacesSlashed}}}{{/if}}
 * {{{implementsTag}}}
 * @extends {{{extendsTag}}}
 * @constructor
 * @public
```

petermuessig (changes requested):
Look fine in general, just a minor suggestion to improve the output.

petermuessig (approved):

PR #1220, opened 2025-05-21
Feat(ui5 middleware serveframework) respect defined libraries in local config

FrankVisuals:
# Why this is useful / needed
We would like to use serveframework also as part of our pull request voting. However, the installation takes quite some time if it installs all the libraries.

I understand that we cannot simply respect the ui5.yaml, as the data is stored for the user (in default behaviour), not for the project. However, when the project is used, it should consider the ui5.yaml.

# Content

This change introduces a new config setting that allows to specify the ui5.yaml that shall be used (defaults to ui5.yaml) in case a different one is used for testing for example.

It also compares the defined libraries to ensure the data is refetched once the libraries list changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ ui5-workspace.yaml
 
 # local markdown files
 TODO.md
+
+# UI5 preload resources
+.ui5-middleware-serveframework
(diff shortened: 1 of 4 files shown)


FrankVisuals:
Will reopen the PR with correct commit messages

PR #1243, opened 2025-07-24
fix(ui5-tooling-modules): aggregation typing is fixed when union type is defined for custom elements

Thodd:
When a union type is defined for a slot in the custom elements manifest, we wrongly only take the first part of the union into account for a control wrapper aggregation.


Thodd:
WIP!

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -37,6 +37,7 @@ const calculateGetterName = (functionName) => {
 // the class name of the base class of all control wrappers
 // corresponds to the "sap.ui.core.webc.WebComponent" class at runtime.
 const UI5_ELEMENT_CLASS_NAME = "UI5Element";
+const WEBC_ASSOCIATION_TYPE = "HTMLElement | string | undefined";
```
petermuessig, line 40: If the order can be guaranteed, I'm ok with this - otherwise I would suggest an array to match. But for a first drop I'm fine

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -341,39 +350,67 @@ class RegistryEntry {
...
+				ui5Type: typeDef.types[0].ui5Type,
+				moduleType: typeDef.types[0].moduleType,
+				isComplexType: true,
+			};
+			return typeDef;
+		}
+
+		const hasNoUnionType = (obj) => obj.dedicatedTypes.length <= 1;
+		// Either single or multiple union type (e.g Array<type1 | type2> or type1 | type2)
+		if (!typeDef.types.every(hasNoUnionType)) {
```
petermuessig, line 391: As discussed yesterday, `hasNoUnionType` and a NOT check afterwards makes it a bit confusing. I personally would not do a double negation.

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
```diff
@@ -341,39 +350,67 @@ class RegistryEntry {
...
+				};
+				return typeDef;
 			}
 		}
-		typeDef.types = typeDef.types.filter((t) => !!t);
+
+		// unpack UI5 type
+		// as UI5 can only handle one type (unlike TS)
+		const refType = typeDef.types[0].dedicatedTypes[0];
+		typeDef.ui5TypeInfo = {
```
petermuessig, line 407: I often saw that you rely on the first type which requires a proper order. I assume that this order is stable and therefore you can properly rely on that. Keep that in mind - potentially a source for an issue in future.

petermuessig (approved):
I'm basically OK with the change - we can go ahead an merge it - just some smaller comments for your change...

packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js
Thodd: see my first comment

PR #1273, opened 2025-09-29
fix(ui5-tooling-modules): ensure default text for enum values in jsdoc

petermuessig:


H4ze (dismissed):
lgtm

petermuessig:
Ahh, need to fix the tests - but there is another small issue I will fix with this change!

PR #1275, opened 2025-09-30
feat(ui5-tooling-modules): support declarative entry point modules

petermuessig:


petermuessig:
@Thodd @H4ze - this change is FYI - I enabled entry point modules in the `ui5.yaml` to get rid of the pseudo modules to define the required dependencies.

PR #1276, opened 2025-09-30
fix(ui5-tooling-modules): change defaults for prettier

petermuessig:


petermuessig:
@H4ze @Thodd - i changed the defaults for prettier but added a `TODO` for all config places - ideally we centralize it and we should also consider using the prettier config of the project if available...

H4ze (commented):
You have been faster providing the change 🫠

PR #1318, opened 2026-02-13
feat(ui5-tooling-modules): allow deeper nested web component classes of the same name

### LINKED ISSUE #1336: ui5-tooling-modules: ShellBarSearch not wrapped as UI5 control when @ui5/webcomponents-fiori >= 2.21.0
### Bug

When using `@ui5/webcomponents-fiori >= 2.21.0` with `addToNamespace: true`, `ShellBarSearch` is never wrapped as a UI5 control in `gen/`. It ends up in `thirdparty/` as raw JS, causing a runtime 404 for `gen/@ui5/webcomponents-fiori/dist/ShellBarSearch.js`.

### Root Cause

`@ui5/webcomponents-fiori@2.21.x`'s `custom-elements-internal.json` exports a default named `ShellBarSearch` from **two modules**:
- `dist/ShellBarSearch.js` — the public web component
- `dist/shellbar/ShellBarSearch.js` — an internal search controller (same export name, different class)

In `WebComponentRegistry.js#parseExports`, the second export **overwrites** `classDef.module` unconditionally:

```js
correspondingClass.module = exp.declaration?.module; // always overwrites
```

This makes the registry alias point to `dist/shellbar/ShellBarSearch.js`. When `lookupWebComponentsClass` resolves the public `dist/ShellBarSearch.js` path, the alias lookup fails → `buildWrapper` is never called → no `gen/` wrapper.

### Fix

Use `??=` so the first (correct) export wins:

```js
correspondingClass.module ??= exp.declaration?.module;
```

### Workaround

`patch-package` patch in use: https://github.com/SAP-samples/uxc-integration/blob/main/patches/ui5-tooling-modules%2B3.35.0.patch

Thodd:
## Summary

Fixes several issues in `ui5-tooling-modules` around web component metadata parsing,
class-name conflicts in deeply nested package layouts, and the recently-broken support
for non-UI5 web component packages.

### Naming conflicts for nested web components
- Allow deeper nested web component classes of the same name by storing classes,
  enums, and interfaces under module-path-based cache keys (`deriveCacheKey`) instead
  of the bare symbol name. Templates use `this.name` rather than `@key`.
- Add correct namespaces for enums so qualified names are emitted consistently.
- Extend type parsing to support `Set<...>` and `Promise<...>` generics.

### Web component metadata parsing
- Make CEM → UI5 metadata derivation more robust (defensive lookups, better fallbacks
  for partially-typed members).
- Fix property-type derivation in `WebComponentRegistry#processMembers`: the existence
  check was reading the never-assigned `typeDef.ui5Type`, so every property fell back
  to `type: "any"` in the generated wrapper. Now reads `typeDef.ui5TypeInfo?.ui5Type`
  and emits the correct UI5 types again (`string`, `boolean`, `sap.ui.core.CSSSize`,
  enum types, etc.).

### Non-UI5 web component packages
- Distinguish UI5 vs. non-UI5 web component packages via a new `isUI5WebComponents`
  flag on the registry entry, instead of skipping non-UI5 packages outright. Wrapper
  classes are now generated for any web component package (e.g.
  `@luigi-project/container`), while UI5-specific chunk emission and base-package
  imports remain gated to actual UI5 packages.
- Update the Luigi container test to match the new package-level output
  (`LuigiContainer` / `LuigiCompoundContainer` are exported as classes, the package
  itself has no `_ui5metadata`).

### Housekeeping
- Regenerate test fixtures and snapshots for `ui5-tooling-modules`.
- Rename a few `webc-registry` fixture files for consistency.
- Relocate pnpm `overrides` from `package.json` to `pnpm-workspace.yaml` (pnpm v10+
  workspace convention) and refresh `pnpm-lock.yaml`.

## Test plan
- [x] `pnpm -r test` for `ui5-tooling-modules`
- [x] Build the `ui5-tsapp` showcase and verify generated wrappers under
      `dist/gen/**` use real UI5 types (no stray `type: "any"`).
- [x] Verify `@luigi-project/container` wrapper classes are generated and usable.
- [x] Smoke-check additional showcases consuming UI5 web components.

Fixes #1336


Thodd:
`classes` and `enums` should work now.
Next steps: `interfaces` --> `this.prefix(...)` must be replaced consistently with fully qualified names.

Thodd:
Tests run green locally.
Filenames for fixtures are now also reflecting the deeper nesting structure of classes.

**Open topics for a new PR**: provide types/enums/interfaces under their correct namespaces in DTS files.

PR #1384, opened 2026-06-16
fix(ui5-tooling-modules): align `@implements` JSDoc with `interfaces` declaration

petermuessig:
## Problem

The generated WebComponent wrappers (`dist/gen/.../<Component>-dbg.js`) emitted an `@implements` JSDoc that did not match the `interfaces` metadata declaration it was meant to describe. Example from `Button-dbg.js`:

```js
// before
* @implements module:ui5/.../gen/ui5/webcomponents.IButton             // ← wrong
...
interfaces: ["ui5....gen.ui5.webcomponents.dist.Button.IButton"]       // ← already correct
```

The `interfaces` declaration is already correct: the interface is named as a sibling of its host class module (`dist.Button.IButton`), mirroring how UI5 references named module exports — cf. the comment on `prefixnsAsModule`. The JSDoc however was hand-built in `#processUI5Interfaces` from `${this.namespace}.${interfaceDef.name}` and dropped the `dist/<Class>` segment entirely.

## Fix

Align the JSDoc with the (correct) interfaces declaration by reusing the precomputed `_ui5QualifiedNameSlashes` so both strings are built from the same source of truth, and ensure that slashed form follows the proper named-export shape (slashes in the module path, dot before the named export):

```js
// after
* @implements module:ui5/.../gen/ui5/webcomponents/dist/Button.IButton
...
interfaces: ["ui5....gen.ui5.webcomponents.dist.Button.IButton"]
```

- `_ui5QualifiedNameSlashes` for interfaces is now computed directly from the class-qualified slashes plus a `.<interfaceName>` suffix instead of blindly converting every dot to a slash.
- `#processUI5Interfaces` now emits `module:${interfaceDef._ui5QualifiedNameSlashes}` so the JSDoc and metadata stay in lockstep by construction.

Also tightened the `@ui5-module-override` tag emitted for interfaces and enums in `JSDocSerializer.js`: it now uses the bare `interfaceDef.name` / `enumDef.name` instead of the cache-key-style `<module>>>...` form, so the override target matches the symbol name. (The tag name itself is still semantically off — the interface/enum is not what gets overridden but its implementing classes / values; flagged with TODOs for a follow-up.)

## Other changes

- Bumped showcases to `@ui5/webcomponents*@^2.23.1`.
- Regenerated `ui5-tooling-modules` test fixtures and snapshots to capture the corrected output.

## Verification

Rebuilt `showcases/ui5-tsapp-webc` and spot-checked Button, Avatar, Breadcrumbs, FormItem, MenuItem — all five wrappers now show **identical** interface references on both sides:

| Wrapper | `@implements` (slashed) | `interfaces:` (dotted) |
|---|---|---|
| Button | `module:.../dist/Button.IButton` | `....dist.Button.IButton` |
| Avatar | `module:.../dist/AvatarGroup.IAvatarGroupItem` | `....dist.AvatarGroup.IAvatarGroupItem` |
| Breadcrumbs | `module:.../dist/ToolbarItem.IToolbarItemContent` | `....dist.ToolbarItem.IToolbarItemContent` |
| FormItem | `module:.../dist/Form.IFormItem` | `....dist.Form.IFormItem` |
| MenuItem | `module:.../dist/Menu.IMenuItem` | `....dist.Menu.IMenuItem` |

All 39 `ui5-tooling-modules` tests pass.

## Out of scope (follow-up)

The augmentation paths in `#ensureDefaults` (e.g. `IFormContent`, `Label`, `ISemanticFormContent`, `IBar` pushed for `ui5-multi-input` / `ui5-shellbar` / Label etc.) push into `ui5metadata.interfaces` *after* `#processUI5Interfaces` has already serialized the JSDoc array. Those entries appear in `interfaces:` but not in `@implements`. That's a separate, pre-existing inconsistency and is **not** addressed by this PR.


Thodd (approved):
LGTM 👍

PR #1398, opened 2026-06-26
fix(ui5-tooling-modules): skip complex-type lookup when typeInfoRef has no module

petermuessig:
### Problem

[WebComponentRegistry#parseComplexType](packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js) resolves complex web-component type references (enums / interfaces / classes) by deriving a cache key from `typeInfoRef.module` + `typeInfoRef.name` via [WebComponentRegistryHelper.deriveCacheKey](packages/ui5-tooling-modules/lib/utils/WebComponentRegistryHelper.js#L46-L53):

```js
deriveCacheKey(keyDef) {
    if (keyDef?.module) {
        return `${keyDef.module}>>${keyDef.name}`;
    }
    return undefined;
}
```

Some web-component custom-elements manifests carry type references that omit `module` (e.g. ambient or library-internal types). With `module` missing the cache key is `undefined`, so:

- Every registry lookup keyed by `undefined` misses.
- The cross-package branch then logs a misleading `Reference package '<pkg>' for complex type '<t>' not found` and proceeds.
- The "global import" fallback (case 1d) returns a `globalImport` entry whose cache lookup was driven by the same `undefined` key, producing an unusable reference.

### Fix

Guard the entire complex-type resolution block on `typeInfoRef.module`. When the module is missing, skip the registry/cross-package/global-import probes entirely and fall through to the existing `any` fallback that the surrounding code already handles:

```js
let complexType;
if (typeInfoRef.module) { // some types are not having a module
    complexType = this.#parseComplexType(typeInfoRef, this);
    // … cross-package + global-import branches …
}
```

No behavior change for type references that do carry a module — those still flow through the existing cases 1b / 1c / 1d unchanged.

### Notes

- Changeset: `patch` for `ui5-tooling-modules`.
- Removes the noisy log line for manifests with module-less type entries.

### DIFF
--- a/.changeset/webc-registry-typeinforef-without-module.md
+++ b/.changeset/webc-registry-typeinforef-without-module.md
@@ -1,0 +1,5 @@
+---
+"ui5-tooling-modules": patch
+---
+
+Guard the complex-type lookup in [WebComponentRegistry#parseComplexType](packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js) against type references that have no `module`. `WebComponentRegistryHelper.deriveCacheKey` returns `undefined` for such references, which previously caused the cross-package lookup and "global import" fallback to misfire (e.g. logging `Reference package '…' for complex type '…' not found` and producing unusable cache entries). When `typeInfoRef.module` is missing, skip the complex-type resolution entirely and fall through to the `any` fallback.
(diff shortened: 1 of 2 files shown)


petermuessig:
@Thodd @H4ze FYI - we should add more logging to the metadata processing - hard to always point out issues in the metadata files which causes side-effects for the web components generation. I'll check what I can do in a follow-up PR.

PR #1404, opened 2026-06-28
fix(ui5-tooling-modules): correct JSDoc generation for enums and defaultValue handling

petermuessig:
## What

Fixes the generation of the JSDoc for enums emitted by `ui5-tooling-modules` so that downstream JSDoc build/validation no longer fails. Also fixes `defaultValue` handling for generated UI5 metadata.

## Why

The generated JSDoc for enum types produced output that the downstream JSDoc build/validation could not resolve (e.g. unresolvable `module:` references, wrong `@ui5-module-override` target). Separately, `defaultValue` strings were being mangled by string-escape stripping and an ad-hoc cast, which broke numeric/boolean/object defaults in the generated metadata.

## How

**JSDoc / enums**

- `JSDocSerializer`: emit `@ui5-module-override` using the derived UI5 class name instead of the raw `enumDef.name`, so the override actually matches the generated module.
- `WebComponentRegistry`:
  - Track both `_ui5QualifiedName` and `_ui5QualifiedNameSlashes` on enum definitions when ingesting the custom-elements metadata.
  - For enum-typed properties, emit `moduleType: module:<slashed-qualified-name>` so the JSDoc validator can resolve them.
  - Propagate `_derivedUi5ClassName` for use by the serializer.

**defaultValue handling**

- Remove `#castDefaultValue` and the string-escape stripping. The raw `propDef.default` from the web component metadata is now passed through directly, which keeps numeric, boolean, and object defaults intact.
- The Handlebars `json` helper now special-cases `defaultValue`:
  - When set to a real value, it is injected as a literal (no surrounding quotes), so values like `true`, `42`, or `{ foo: 1 }` stay as literals in the generated code.
  - When set to the string `"undefined"`, the property is dropped from the JSON output.

## Files

- `packages/ui5-tooling-modules/lib/utils/HandlebarsHelper.js`
- `packages/ui5-tooling-modules/lib/utils/JSDocSerializer.js`
- `packages/ui5-tooling-modules/lib/utils/WebComponentRegistry.js`


petermuessig:
@Thodd @H4ze - FYI - this fixes the inconsistencies in the enum names and makes them working in JSDoc +  I fixed the defaultValue handling which was corrupt - there was a reason for the quoting of the values... :-)

PR #1406, opened 2026-06-29
fix(ui5-tooling-modules): sync defaultValue handling between rollup plugin and Handlebars helper

petermuessig:
## Summary

Aligns the `defaultValue` serialization in [rollup-plugin-webcomponents.js](packages/ui5-tooling-modules/lib/rollup-plugin-webcomponents.js) with the existing logic in [HandlebarsHelper.js](packages/ui5-tooling-modules/lib/utils/HandlebarsHelper.js), so the generated metadata is identical regardless of which code path produces it.

## Changes

- **[rollup-plugin-webcomponents.js](packages/ui5-tooling-modules/lib/rollup-plugin-webcomponents.js)**: Pass a replacer to `JSON.stringify` for the metadata object that handles `defaultValue` specially:
  - `undefined` / `"undefined"` → property dropped
  - `""` → emitted as empty string
  - otherwise → parsed via `JSON.parse` so booleans, numbers, objects and quoted strings come out as proper JSON values (e.g. `true` instead of `"true"`, `"Button"` instead of `'"Button"'`)
  - non-parseable values are dropped with a warning
- **[HandlebarsHelper.js](packages/ui5-tooling-modules/lib/utils/HandlebarsHelper.js)**: Added a sync note pointing at the rollup plugin so the two stay aligned.
- **Snapshots**: Updated `CheckBox`, `Button`, `Panel` and `LuigiContainer` snapshots to reflect the corrected metadata output.
- **Changeset**: Added a `patch` changeset for `ui5-tooling-modules`.

## Why

Previously, `defaultValue` was serialized as a JSON string (e.g. `"true"`, `'"Button"'`, `"undefined"`) which made the metadata inconsistent and invalid for the UI5 runtime. The Handlebars-based path already handled this; the rollup plugin path did not. Both now produce the same shape.

### DIFF
--- a/.changeset/fix-ui5-tooling-modules-defaultvalue-sync.md
+++ b/.changeset/fix-ui5-tooling-modules-defaultvalue-sync.md
@@ -1,0 +1,9 @@
+---
+"ui5-tooling-modules": patch
+---
+
+Sync `defaultValue` serialization between `rollup-plugin-webcomponents` and the Handlebars `json` helper so the generated UI5 metadata is identical regardless of code path.
+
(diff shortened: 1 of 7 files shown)


petermuessig:
Final one now. Ideally in future we harmonize the different tracks in the code to avoid issues...

PR #1424, opened 2026-08-10
feat(ui5-tooling-modules): add create-webc-controls CLI to generate UI5 control wrappers

Thodd:
Tooling for creating UI5 wrapper controls for native HTML elements.


petermuessig (approved):
Free from side effects... So it's good to go...

PR #1427, opened 2026-08-13
fix(ui5-tooling-modules): fix generated Web Component control wrappers

H4ze:
## Summary

Three fixes and enhancements for the \`create-webc-controls\` CLI that generates UI5 control wrappers from a Custom Elements Manifest (CEM):

- **Keep quotes on metadata property keys** (\`d0cd452\`): Generated \`_ui5metadata\` objects with keys containing special characters (hyphens, dots) were losing their quotes, producing invalid JavaScript. Fixed by setting \`quoteProps: "preserve"\` in the Prettier formatting pass.

- **Support enum member value divergence** (\`1cf5b60\`): When a Web Component enum member has a different HTML attribute value than its JavaScript identifier name (e.g. identifier \`Num1\` for the HTML value \`"1"\`), \`WebComponentRegistry\` now honours the CEM \`value\` overlay field so the generated \`registerEnum\` call maps the correct runtime value. Updated \`UI5Package.hbs\` to render \`{{this.value}}\` instead of \`{{this.name}}\`.

- **Add \`libraryMode\` option for UI5 library generation** (\`ba51d0a\`, \`7f67378\`): Adds a \`libraryMode\` flag to \`generateControls()\` for generating a proper \`Library.init()\`-based \`library.js\` (required for real UI5 libraries) instead of the standalone \`pkg["_ui5metadata"]\` package module. The output path is derived automatically from \`outputDir\` and the UI5 namespace (e.g. \`sap.html\` → \`sap/html/library.js\`), so no extra parameter is needed — callers only set \`libraryMode: true\`. Uses the new \`UI5Library.hbs\` template which emits \`Library.init({...})\` with full \`controls\`/\`types\`/\`interfaces\` arrays and all \`registerEnum\` calls. The existing \`buildPackage()\` / \`UI5Package.hbs\` path is unchanged.

## Test plan

- [ ] Run \`create-webc-controls\` on a CEM with quoted metadata keys — verify generated JS is syntactically valid
- [ ] Run \`create-webc-controls\` on a CEM that includes an enum with diverging identifier/value (e.g. \`@ui5/html\` \`ListType\` where \`Num1 = "1"\`) — verify \`registerEnum\` output maps \`"Num1": "1"\` correctly
- [ ] Run \`generateControls({ ..., libraryMode: true })\` on a CEM with enums — verify the generated \`library.js\` contains \`Library.init\` with full control/type arrays and all \`registerEnum\` calls at \`<outputDir>/<ui5Namespace>/library.js\`
- [ ] Verify that running without \`libraryMode\` still produces the same standalone package module as before


Thodd:
The many fixtures are because of the `value` that is now serialized for enums.

petermuessig (approved):
LGTM

PR #1433, opened 2026-08-21
feat(ui5-tooling-modules): capabilities for control wrapper and library generation via cli

H4ze:
## Summary

This PR adds extended capabilities to generate control wrappers (and to a lesser extend libraries) through the `createControls` CLI.

As of now mainly intended to be used for the generation of the a library in OpenUI5.

The following features are introduced

### 1. Class Name Generation

- Replace generic `WebComponentBaseClass` with the actual base class name (e.g. `WebComponent`, `HTMLElementBase`) in generated AMD wrapper files
- Replace generic `WrapperClass` with the actual UI5 class simple name (e.g. `CustomAlertButton`, `LuigiContainer`) in generated AMD wrapper files
- Applied to both the Rollup plugin (`rollup-plugin-webcomponents.js`) and the CLI code path (`cli/createControls.js`)

#### Motivation

The generated wrappers previously used `WebComponentBaseClass` and `WrapperClass` as variable names, making the generated code harder to read and debug. Using the actual class names makes the generated output self-documenting.

### 2. Enum Generation

- Enums (e.g. for HTML support) must be correctly nested within a UI5 library.js
  - Only active for the `createControls` CLI path
  - Needs deeper triage for the main web-components generation path (specifically regarding compatibility)
- Enums are now correctly sub-namespaced with dot-syntax instead of bracket syntax.
  - e.g. `thisLib.enums.MyEnum` instead of `thisLib.enums["MyEnum"]`
  - This is done to work around a quirk in the OpenUI5 jsdoc builder plugin and the [UI5/typescript/dts-generator](https://github.com/UI5/typescript/tree/main/packages/dts-generator) which picks up the bracket syntax as "exports" and then downstream causes wrongly generated library DTS files.
- Enums now allow for (correctly) harmonized key<>value pairs.
  - This is important as deviating key<>value pairs violate the convention in UI5
- Enums now allow non-identifier keys
  - The DOM itself supports attribute value that are invalid as identifiers in ES
  - To account for this we now quote all enum keys and if needed access them via bracket syntax instead of dot syntax
  - **Sample**: The `<input>` element has a `type` attribute which supports a fixed list of types, of which some are using a hyphen (`-`), e.g. `datetime-local`. Refer to [<input> types @ MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#input_types). Many more of this kind exist throught the DOM.
  - **Sample 2**: The `<ol>` element has a `type` attribute which accepts a number (`1`) as value. The number `1` is not a valid enum key in TypeScript (neither quoted nor unquoted). Refer to [ol#type @ MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ol#type).

#### Motivation

As of now enum namespaces are not 100% correct.
Fixing them everywhere would lead to publicly noticeable change in the TS support.
For now we fix them in the CLI variant that just generates controls and `libray.js`.

### 3. Necessary Fixes

The 
Native (non-UI5Element) wrappers dropped the side-effect import of the native
custom element implementation when the superclass resolved to a truthy
fallback, so customElements.define() never ran at runtime. Import the native
impl whenever it exists and isn't already the package/namespace import slot.

Also add a SubCustomAlertButton→CustomAlertButton inheritance test and make the
test harness's mocked WebComponent.extend return an extendable class.

#### Why fix it in this PR?

The SubCustomAlertButton inheritance fixture @petermuessig asked for flips CustomAlertButton's resolution, and without the fix in this PR we ship a wrapper whose custom element never gets registered at runtime.

This is actually a latent bug (already in main), but became visible through the added test and the changes to `webcBaseClassName` and `namespaceImport`. Those changes are needed for the correct import naming, but then incidentally make the pre-existing issue visible. So we should fix it here instead of separately, otherwise this PR would introduce the issue to consumers.


petermuessig:
I didn't see a test for the custom base classes - did you verify it that it works for them as well? Currently all tests only relate to WebComponent as base class.

petermuessig:
**Bingo! I found a real vulnerability!** 

## ⚠️ **CRITICAL ISSUE: Forbidden Characters in Variable Names**

The code **does NOT validate** that `ui5ClassSimpleName` is a valid JavaScript identifier. If a component name contains forbidden characters in the last segment (after the final dot), the generated code will be **syntactically invalid**.

### Problematic Cases:
- `some.package.My-Component` → `My-Component` ❌ (hyphen is invalid)
- `some.package.My Component` → `My Component` ❌ (space is invalid)
- `some.package.123Component` → `123Component` ❌ (starts with number)
- `some.package.My@Component` → `My@Component` ❌ (@ is invalid)

### Current Code (Vulnerable):
```javascript
const ui5ClassSimpleName = ui5ClassName.split(".").pop();
```

### Recommendation:
Add validation to ensure the name is a valid JavaScript identifier, or sanitize it:

```javascript
const ui5ClassSimpleName = ui5ClassName.split(".").pop();
// Validate it's a valid identifier
if (!/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(ui5ClassSimpleName)) {
  throw new Error(`Invalid identifier: "${ui5ClassSimpleName}" from "${ui5ClassName}"`);
}
```

Or sanitize it:
```javascript
const ui5ClassSimpleName = ui5ClassName.split(".").pop().replace(/[^a-zA-Z0-9_$]/g, "_");
```

**The current test snapshots don't catch this because they only use valid names.** This is a latent bug that could cause failures with certain component names.

Thodd:
Class names are now derived as valid JS identifiers and I added a helper function so that both places don't drift.

Thodd:
> **Bingo! I found a real vulnerability!**
> 
> ## ⚠️ **CRITICAL ISSUE: Forbidden Characters in Variable Names**
> The code **does NOT validate** that `ui5ClassSimpleName` is a valid JavaScript identifier. If a component name contains forbidden characters in the last segment (after the final dot), the generated code will be **syntactically invalid**.
>  [...]

Fixed in the meantime. We use the same regex for Identifier detection in our library generator in OpenUI5 as well.

Thodd:
> I didn't see a test for the custom base classes - did you verify it that it works for them as well? Currently all tests only relate to WebComponent as base class.

Do you think an additional test is necessary? I can add one.

Still technically it works nicely, but at the moment you can only observe it internally in our new library for now :)

petermuessig:
Tests would be great - but we can also add them later if you want to continue now.

petermuessig (dismissed):
You can go ahead, merge and release - as said tests can follow separately...

Thodd:
> You can go ahead, merge and release - as said tests can follow separately...

Alright thanks.
There is a security topic I still want to cover (URL validation) so maybe I'll need to make another commit.
If I do, I'll add some tests in addition too.

packages/ui5-tooling-modules/test/__fixtures__/@ui5/webcomponents-fiori/classes/ui5-webcomponents-fiori-dist-NavigationMenu.json
```diff
@@ -1 +1 @@
...
...Control","types":[{"origType":"HTMLElement","multiple":false,"dedicatedTypes":[{"dtsType":"Control","packageName":"sap/ui/core/Control","moduleType":"module:sap/ui/core/Control","ui5Type":"sap.ui.core.Control","isClass":true}]}],"dtsParamDescription":"The `ui5-menu-item` that triggers opening of the sub-menu or undefined when fired upon root menu opening."}}},"close":{"allowPreventDefault":false,"enableEventBubbling":false,"parameters":{}},"itemClick":{"allowPreventDefault":true,"enableEventBubbling":false,"parameters":{"item":{"type":"sap.ui.core.Control","types":[{"origType":"HTMLElement","multiple":false,"dedicatedTypes":[{"dtsType":"Control","packageName":"sap/ui/core/Control","moduleType":"module:sap/ui/core/Control","ui5Type":"sap.ui.core.Control","isClass":true}]}],"dtsParamDescription":"The currently clicked menu item."},"text":{"type":"string","types":[{"origType":"string","multiple":false,"dedicatedTypes":[{"dtsType":"string","ui5Type":"string"}]}],"dtsParamDescription":"The text of the currently clicked menu item."}}},"onOpen":{"allowPreventDefault":false,"enableEventBubbling":true,"parameters":{},"mapping":"open"}},"getters":[],"methods":[],"defaultAggregation":"items"}
```
Thodd, line 1: hard to see in the diff, but the `opener` now has the correct typing as well.

codeworrior: Yes, I've seen that. I just wondered whether we want to keep it like that or suppress type "sap.ui.core.Control" as it is the default in the runtime.

packages/ui5-tooling-modules/cli/createControls.js
```diff
@@ -198,13 +198,65 @@ function buildLibrary({ package: pkg, template, outputDir }) {
...
+
 	const code = template({
 		metadata,
-		hasEnums: Object.keys(pkg.enums).length > 0,
-		enums: pkg.enums,
+		libraryJsDoc,
+		hasEnums: enums.length > 0,
+		enums,
+		enumNamespaces,
 		dependencies: pkg.dependencies?.map((dep) => dep),
```
codeworrior, line 262: AFAICS, a RegistryEntry in the WCR has no property `dependencies`. Therefore, no further dependencies are generated here, esp. not the one required for lib dependency "sap.ui.core".

Maybe the code rather wanted to access `metadataObject.dependencies` and convert each entry to a library module (e.g. "sap/ui/core/library")?

codeworrior: I have to partially correct myself. While the WCR does not manage any dependencies, the rollup-plugin-webcomponents does and attaches them to the registry entry. But for the generateControls API, this seems not to happen.

Thodd: Ack.

packages/ui5-tooling-modules/test/__snap__/56bfa72a/@luigi-project/container/LuigiContainer.js
```diff
@@ -2,11 +2,11 @@
  * ${copyright}
  */
 sap.ui.define(
-  ["sap/ui/core/webc/WebComponent", "../../@luigi-project/container"],
-  function (WebComponentBaseClass) {
+  ["sap/ui/core/webc/WebComponent", "../../LuigiContainer"],
```
petermuessig, line 5: It's now like this because the class is not a UI5Element but it has a superclass (which was the indicator before). But the behavior is correct now and also works!

petermuessig (approved):
LGTM

